Issue #12, opened 2016-02-23
Provide some utilities for other rules


sindresorhus:
Some of the things in https://github.com/sindresorhus/eslint-plugin-ava/pull/10 could be useful for other rules and should be extracted into `util.js`.

twada:
@sindresorhus Okay I'll extract common structure between existing `test-ended` and `prefer-power-assert` rules.

PR #13, opened 2016-02-23
add test-title rule (fixes #8)

### LINKED ISSUE #8: `test-title` rule
Require tests to have a title. Should have an option to enable an exception for when there is only one test.

jfmengels:
Adds test-title rule (fixes #8)

Let me know if I need to change anything ;)

### DIFF
--- a/docs/rules/test-title.md
+++ b/docs/rules/test-title.md
@@ -1,0 +1,37 @@
+# Ensure tests have a title
+
+Tests should have a title.
+
+
+## Fail
+
+```js
+const test = require('ava');
+
+test(t => {
+	t.pass();
+	t.end();
+});
+```
+
+
+## Pass
+
+```js
+const test = require('ava');
+
+test('my test name', t => {
+	t.pass();
+	t.end();
+});
+```
+
+## Options
+
+The rule takes one option, a string, which could be either `"always"` or `"if-multiple"`. The default is `"always"`. If the option is set to `"if-multiple"`, the rule will only trigger if there are multiple tests in a file.
+
+You can set the option in configuration like this:
+
+```js
+"test-title": [2, "always"]
+```
--- a/rules/test-title.js
+++ b/rules/test-title.js
@@ -1,0 +1,31 @@
+'use strict';
+var createAvaRule = require('../create-ava-rule');
+
+/* eslint quote-props: [2, "as-needed"] */
+module.exports = function (context) {
+	var ifMultiple = context.options[0] === 'if-multiple';
+	var testCount = 0;
+
+	var ava = createAvaRule();
+	return ava.merge({
+		CallExpression: function (node) {
+			if (!ava.isTestFile || ava.currentTestNode !== node) {
+				return;
+			}
+
+			testCount++;
+			var hasNoTitle = node.arguments.length !== 2;
+			var isOverThreshold = !ifMultiple || testCount > 1;
+			if (hasNoTitle && isOverThreshold) {
+				context.report(node, 'Test should have a title.');
+			}
+		},
+		'Program.exit': function () {
+			testCount = 0;
+		}
+	});
+};
+
+module.exports.schema = [{
+	enum: ['always', 'if-multiple']
+}];
--- a/test/test-title.js
+++ b/test/test-title.js
@@ -1,0 +1,53 @@
+import test from 'ava';
+import {RuleTester} from 'eslint';
+import rule from '../rules/test-title';
+
+const ruleTester = new RuleTester({
+	env: {
+		es6: true
+	}
+});
+
+const errors = [{ruleId: 'test-title'}];
+const header = `const test = require('ava');\n`;
+
+test(() => {
+	ruleTester.run('test-title', rule, {
+		valid: [
+			header + 'test("my test name", function (t) { t.pass(); t.end(); });',
+			header + 'test(`my test name`, function (t) { t.pass(); t.end(); });',
+			header + 'test(\'my test name\', function (t) { t.pass(); t.end(); });',
+			header + 'test.cb("my test name", t => { t.pass(); t.end(); });',
+			{
+				code: header + 'test(function (t) { t.pass(); t.end(); });',
+				options: ['if-multiple']
+			},
+			header + 'notTest(t => { t.pass(); t.end(); });',
+			// shouldn't be triggered since it's not a test file
+			'test(t => {});'
+		],
+		invalid: [
+			{
+				code: header + 'test(function (t) { t.pass(); t.end(); });',
+				errors
+			},
+			{
+				code: header + 'test.cb(function (t) { t.pass(); t.end(); });',
+				errors
+			},
+			{
+				code: header + 'test.cb.skip(function (t) { t.pass(); t.end(); });',
+				errors
+			},
+			{
+				code: header + 'test(t => { t.pass(); t.end(); });',
+				errors
+			},
+			{
+				code: header + 'test(function (t) { t.pass(); t.end(); }); test(function (t) { t.pass(); t.end(); });',
+				options: ['if-multiple'],
+				errors
+			}
+		]
+	});
+});
--- a/index.js
+++ b/index.js
@@ -4,7 +4,8 @@ module.exports = {
 	rules: {
 		'no-skip-test': require('./rules/no-skip-test'),
 		'prefer-power-assert': require('./rules/prefer-power-assert'),
-		'test-ended': require('./rules/test-ended')
+		'test-ended': require('./rules/test-ended'),
+		'test-title': require('./rules/test-title')
 	},
 	configs: {
 		recommended: {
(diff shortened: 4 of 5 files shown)


docs/rules/test-title.md
```diff
@@ -0,0 +1,35 @@
...
+
+Tests should have a title.
+
+
+## Fail
+
+```js
+const test = require('ava');
+
+test((t) => {
```
sindresorhus:
```js
test(t => {
```

docs/rules/test-title.md
```diff
@@ -0,0 +1,35 @@
...
+});
+```
+
+
+## Pass
+
+```js
+const test = require('ava');
+
+test('my test name', (t) => {
```
sindresorhus: same as https://github.com/sindresorhus/eslint-plugin-ava/pull/13/files#r54228577

docs/rules/test-title.md
```diff
@@ -0,0 +1,35 @@
...
+});
+```
+
+## Options
+
+The rule takes one option, a string, which could be either `"always"` or `"if-multiple"`. The default is `"always"`. If the option is set to `"if-multiple"`, the rule will only trigger if there are multiple tests in a file.
+
+You can set the option in configuration like this:
+
+"test-title": [2, "always"]
```
sindresorhus: This should be in a code block

jfmengels: Isn't the case in the `eslint` project rules, but sure, added it.

jfmengels:
Updated! (do not bother making the same comments on my other PRs, will apply them right away)

rules/test-title.js
```diff
@@ -0,0 +1,32 @@
...
+			if (isTestFile) {
+				var args = node.arguments;
+				var callee = node.callee;
+				var name = callee.type === 'MemberExpression' ? callee.object.name : callee.name;
+				if (name === 'test') {
+					testCount++;
+					var hasNoTitle = args.length !== 2 || literalTypes.indexOf(args[0].type) === -1;
+					var isOverThreshold = !ifMultiple || testCount > 1;
+					if (hasNoTitle && isOverThreshold) {
+						context.report(node, 'Test should specify a title.');
```
sindresorhus: > Test should have a title.

jfmengels: Done

sindresorhus:
Hah, you're fast :P

Doesn't look like this support chaining? `test.skip.cb()`. The plan is to extract https://github.com/sindresorhus/eslint-plugin-ava/pull/10/files#diff-5bdffee573ffb3ac8e41b63009aec3daR53, see https://github.com/sindresorhus/eslint-plugin-ava/issues/12.

sindresorhus:
@twada Would you be able to help me review these pull requests? You're the local AST expert ;)

sindresorhus:
// @novemberborn In case you have some interest in helping out reviewing. No worries if not.

jfmengels:
You're right, it doesn't support `test.skip.cb`. Should I create extract `isTestFunctionCall` from #12 in this PR, or make/wait for a new PR to do this?

sindresorhus:
Do it in this PR as it's needed here.

test/test-title.js
```diff
@@ -0,0 +1,49 @@
...
+			{
+				code: header + 'test(function (t) { t.pass(); t.end(); });',
+				errors
+			},
+			{
+				code: header + 'test.cb(function (t) { t.pass(); t.end(); });',
+				errors
+			},
+			{
+				code: header + 'test((t) => { t.pass(); t.end(); });',
```
sindresorhus: `(t)` => `t`

jfmengels:
Ok, will do that (a bit later though ;) ).
I will move the `ImportDeclaration`/`VariableDeclarator` parts too as they improve the "is this a test file" detection (and will update my other PRs later too then. Maybe I should have made only 1 PR for 1 rule to begin with ;)  ).

sindresorhus:
> Maybe I should have made only 1 PR for 1 rule to begin with ;) ).

That would indeed have made easier for you, but would also have slowed down the feedback loop considerably, so I think this a good middle-ground. The merge conflicts should be easy to resolve.

twada:
> @twada Would you be able to help me review these pull requests? You're the local AST expert ;)

Okay ;)

rules/test-title.js
```diff
@@ -0,0 +1,32 @@
...
+				return;
+			}
+
+			if (isTestFile) {
+				var args = node.arguments;
+				var callee = node.callee;
+				var name = callee.type === 'MemberExpression' ? callee.object.name : callee.name;
+				if (name === 'test') {
+					testCount++;
+					var hasNoTitle = args.length !== 2 || literalTypes.indexOf(args[0].type) === -1;
```
twada: Test title could be any Expression (`Identifier`, `BinaryExpression`, `CallExpression`...and so on).
```js
const testName = '...';

// Identifier
test(testName, t => {
});

// BinaryExpression
test('prefix_' + testName + '_suffix', t => {
});
```
So you should just count `args.length` here.

sindresorhus: To complicate it further, the title could be just a function name:

```js
test(function foo() {

});
```

`foo` is the used test title in the above.

jfmengels: Ooh, I didn't know that. So in this case, `foo` would be the test name, and that would be ok. But `t => {}` and `function() {}` would not.
Maybe worth making an option forbidding the use of `function foo() {}` to specify the title name?

sindresorhus: > Maybe worth making an option forbidding the use of function foo() {} to specify the title name?

Why? Like, what would be the gain of doing so?

jfmengels: I think that some users (me first) would like a rule enforcing a "written" rule, that describes the tested behavior in a "string" format, rather than in a function name.

```js
// BAD
test(function shouldIncrementIfTypeIsINCREMENT(t) { ... });
test(function shouldDecrementIfTypeIsDECREMENT(t) { ... });

// GOOD
test('should increment value if type is INCREMENT', function(t) { ... });
test('should decrement value if type is DECREMENT', function(t) { ... });
```

sindresorhus: Yeah, it seemed nice when I added it, but it's not very useful and creates inconsistency. Maybe we should just deprecate it in AVA (?). Hmm. Maybe hold off on supporting this pattern for now, until we decide on what to do about it. I don't want it holding this up.

jfmengels: Yeah, it sounds like a cool trick at first, but not sure I would really want to see this in my codebase when strings are available and more readable. Will not implement it in this PR, but will add @twada's suggestion.

Don't know about deprecating it in AVA, why not.

jfmengels:
Updated. I moved a few things out of `prefer-power-assert` and into `utils`, (updated the existing rules too). I added some tests, but I don't feel comfortable using made-up nodes like I just did. An option would be to use a tool like `recast` to take a string and convert it to a node. What do you guys think?

(Also, didn't add tests for `isAvaImportDeclaration` as with made-up nodes, you'd pretty much have to copy-paste the implementation)

jfmengels:
Maybe put this on hold until #20 gets merged? ^^'

sindresorhus:
#20 got merged. Sorry about the churn... But now we have a good base to build on :)

jfmengels:
Don't worry about it ;)
Have removed my refactoring and used @twada's changes. It makes things a bit easier, thanks for that!

rules/test-title.js
```diff
@@ -0,0 +1,28 @@
...
+
+/* eslint quote-props: [2, "as-needed"] */
+module.exports = function (context) {
+	var ifMultiple = context.options[0] === 'if-multiple';
+	var testCount = 0;
+
+	var ava = createAvaRule();
+	return ava.merge({
+		CallExpression: function (node) {
+			if (!ava.isTestFile || !util.isTestFunctionCall(node.callee)) {
```
sindresorhus: `if (!ava.isTestFile || !ava.currentTestNode) {`

Per https://github.com/sindresorhus/eslint-plugin-ava/pull/20/files#diff-a683eee4dfe24b3d7070e7482ded2411R10

rules/test-title.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+var util = require('../util');
+var createAvaRule = require('../create-ava-rule');
+
+/* eslint quote-props: [2, "as-needed"] */
+module.exports = function (context) {
+	var ifMultiple = context.options[0] === 'if-multiple';
```
sindresorhus: Should we maybe use a options schema to prevent users from doing mistakes in the config? http://eslint.org/docs/developer-guide/working-with-rules#options-schemas

rules/test-title.js
twada: Better to check node equality.
```js
if (!ava.isTestFile || ava.currentTestNode !== node) {
```
Since `if (!ava.isTestFile || !ava.currentTestNode) {` matches all CallExpressions inside `test` function.

rules/test-title.js
jfmengels: :+1:

jfmengels:
Updated!

sindresorhus:
Woot. Looks good :)

jfmengels:
Yeah! Will start creating the PRs for the other ones :)

PR #11, opened 2016-02-21
Add plenty of rules (#2, #3, #5, #6, #7 and #8)

jfmengels:
I implemented the rules in issues #2, #3, #5, #6, #7 and #8.

I'm not that familiar with Ava (but pretty excited about it, so I felt pretty motivated to do this), so let me     know if I've got anything wrong, if there's anything to fix, or if you prefer this being split into multiple PRs.

Also, if someone can tell me why the use of `const` makes the tests fail in Node 0.10/0.12 but is ok on master where I also see `const`...

PR #14, opened 2016-02-23
add no-skip-test rule (fixes #3)

### LINKED ISSUE #3: `no-skip-test` rule
It's easy to do make a test skipped with `test.skip()` and then forget about it. It's visible in the results, but still easily missed. Some projects/companies might want to prevent such tests from entering the CI.

jfmengels:
Adds no-skip-test rule (fixes #3)

Let me know if I need to change anything ;)

One this gets merged, I'll submit the PRs for all, or most of, the rules I originally submitted in #11.

### DIFF
--- a/docs/rules/no-skip-test.md
+++ b/docs/rules/no-skip-test.md
@@ -1,0 +1,27 @@
+# Ensure no tests are skipped
+
+It's easy to make a test skipped with test.skip() and then forget about it. It's visible in the results, but still easily missed.
+
+
+## Fail
+
+```js
+const test = require('ava');
+
+test.skip('some title', t => {
+	t.pass();
+	t.end();
+});
+```
+
+
+## Pass
+
+```js
+const test = require('ava');
+
+test('some title', t => {
+	t.pass();
+	t.end();
+});
+```
--- a/rules/no-skip-test.js
+++ b/rules/no-skip-test.js
@@ -1,0 +1,13 @@
+'use strict';
+var createAvaRule = require('../create-ava-rule');
+
+module.exports = function (context) {
+	var ava = createAvaRule();
+	return ava.merge({
+		CallExpression: function (node) {
+			if (ava.isTestFile && ava.currentTestNode === node && ava.hasTestModifier('skip')) {
+				context.report(node, 'No tests should be skipped.');
+			}
+		}
+	});
+};
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@
 
 module.exports = {
 	rules: {
+		'no-skip-test': require('./rules/no-skip-test'),
 		'prefer-power-assert': require('./rules/prefer-power-assert'),
 		'test-ended': require('./rules/test-ended')
 	},
@@ -15,6 +16,7 @@ module.exports = {
 				sourceType: 'module'
 			},
 			rules: {
+				'no-skip-test': 0,
 				'prefer-power-assert': 0,
 				'test-ended': 2
 			}
--- a/readme.md
+++ b/readme.md
@@ -29,6 +29,7 @@ Configure it in `package.json`.
 			"ava"
 		],
 		"rules": {
+			"ava/no-skip-test": 2,
 			"ava/prefer-power-assert": 0,
 			"ava/test-ended": 2
 		}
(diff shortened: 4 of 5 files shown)


docs/rules/no-skip-test.md
```diff
@@ -0,0 +1,27 @@
+# Ensure no tests are skipped
+
+It's easy to make a test skipped with test.skip() and then forget about it. It's visible in the results, but still easily missed.
```
sindresorhus: Maybe mention that this can potentially be disastrous if you deploy code and your whole test suite is pretty much not validating anything.

jfmengels: Definitely see that for `no-only-test`, but sure, will add that here too.

sindresorhus: Oh, never mind. My brain was in `no-only-test`.

sindresorhus:
LGTM

jfmengels:
Updated to use `createAvaRule`

rules/no-skip-test.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+var util = require('../util');
+var createAvaRule = require('../create-ava-rule');
+
+module.exports = function (context) {
+	var ava = createAvaRule();
+	return ava.merge({
+		MemberExpression: function (node) {
+			if (ava.isTestFile && node.property.name === 'skip' && util.isCallFromTestObject(node)) {
```
sindresorhus: Couldn't this use `ava.hasTestModifier('skip')`?

https://github.com/sindresorhus/eslint-plugin-ava/pull/20/files#diff-a683eee4dfe24b3d7070e7482ded2411R10

sindresorhus:
// @twada review?

rules/no-skip-test.js
twada: This is not wrong since @jfmengels focused on `MemberExpression` then traverse it up.

However if you want to use existing properties and methods, this could be written as follows.
```js
	return ava.merge({
		CallExpression: function (node) {
			if (ava.isTestFile && ava.currentTestNode === node && ava.hasTestModifier('skip')) {
				context.report(node, 'No tests should be skipped.');
			}
		}
	});
```

`ava.currentTestNode === node` means "on entering `test` method". Fires exactly once for each `test` method.

sindresorhus: Yup, that looks better.

jfmengels: Ha! I thought there would be some kind of race condition if I put that in a CallExpression, but it's well written :D (though I think it'll break for nested tests. Don't know if that is currently possible with Ava now)
I'll add that anyway some time today!

sindresorhus: > though I think it'll break for nested tests

AVA doesn't support nested tests.

jfmengels: OK, that simplifies things. Saw a thread about integrating it, but I didn't know how Ava currently treated them (error, badly formatted message, crash...)

jfmengels: Hum, tried using `ava.currentTestNode === node` but it doesn't work here. They don't reference the same thing (`node` references the `MemberExpression`, while `currentTestNode` references the `CallExpression`, aka the parent of `node`).

I tried using `ava.isTestFile && ava.currentTestNode && ava.hasTestModifier('skip')` instead, but then I get two errors instead of one for `test.skip(...)` as it gets triggered for two of the nodes it passes through, and 4 for `test.skip.cb(...)`.
So I'm sticking to this implementation except if someone has a better idea.

sindresorhus: @jfmengels See @twada's example. You need to use a `CallExpression` hook instead.

jfmengels: Ha, missed that part, my bad. Yeah that works. I'll make the change.

jfmengels:
@sindresorhus @twada Updated using your suggestions ;)

sindresorhus:
Perfect! :)

jfmengels:
\o/
Thank you @sindresorhus and @twada, hope this didn't take too much of your time :)

twada:
@jfmengels @sindresorhus :tada:

PR #18, opened 2016-02-26
Update to ESLint 2.0.0

### LINKED ISSUE #4: Add recommended rule settings
This will be possible when ESLint 2.0.0 is released.

See: https://github.com/eslint/eslint/issues/3659

### LINKED ISSUE #17: Upgrade to ESLint 2
http://eslint.org/docs/user-guide/migrating-to-2.0.0

wyze:
In the new recommended config I set up some defaults for environment and parser options. I noticed that you can also use `const ava = require('ava')` and not need `sourceType: module`. But all of the examples of AVA use the `import` syntax so I think this should be in the recommended configuration.

Feedback is definitely welcome.

Fixes #4
Fixes #17

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,8 +5,19 @@ module.exports = {
 		'prefer-power-assert': require('./rules/prefer-power-assert'),
 		'test-ended': require('./rules/test-ended')
 	},
-	rulesConfig: {
-		'prefer-power-assert': 0,
-		'test-ended': 0
+	configs: {
+		recommended: {
+			env: {
+				es6: true
+			},
+			parserOptions: {
+				ecmaVersion: 6,
+				sourceType: 'module'
+			},
+			rules: {
+				'prefer-power-assert': 0,
+				'test-ended': 2
+			}
+		}
 	}
 };
--- a/package.json
+++ b/package.json
@@ -50,11 +50,11 @@
   },
   "devDependencies": {
     "ava": "*",
-    "eslint": "^1.10.3",
+    "eslint": "^2.2.0",
     "js-combinatorics": "^0.5.0",
     "xo": "*"
   },
   "peerDependencies": {
-    "eslint": ">=1.10.0"
+    "eslint": ">=2"
   }
 }
--- a/readme.md
+++ b/readme.md
@@ -12,12 +12,19 @@ $ npm install --save-dev eslint eslint-plugin-ava
 
 ## Usage
 
-Configure it in package.json.
+Configure it in `package.json`.
 
 ```json
 {
 	"name": "my-awesome-project",
 	"eslintConfig": {
+		"env": {
+			"es6": true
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -50,11 +50,11 @@
...
   "devDependencies": {
     "ava": "*",
-    "eslint": "^1.10.3",
+    "eslint": "^2.2.0",
     "js-combinatorics": "^0.5.0",
     "xo": "*"
   },
   "peerDependencies": {
-    "eslint": ">=1.10.0"
+    "eslint": "^2.0.0"
```
sindresorhus:
```
"eslint": ">=2"
```

We don't want peerDependencies to be too strict.

index.js
```diff
@@ -5,8 +5,19 @@ module.exports = {
...
+			env: {
+				es6: true
+			},
+			parserOptions: {
+				ecmaVersion: 6,
+				sourceType: 'module'
+			},
+			rules: {
+				'prefer-power-assert': 0,
+				'test-ended': 0
```
sindresorhus:
```
'test-ended': 2
```

There's no reason not to want this one.

package.json
kentcdodds: :+1:

readme.md
```diff
@@ -38,6 +45,29 @@ The rules will only activate in test files.
 - [prefer-power-assert](docs/rules/prefer-power-assert.md) - Allow only use of the asserts that have no [power-assert](https://github.com/power-assert-js/power-assert) alternative.
 
 
+## Recommended configuration
+
+This plugin exports a `recommended` configuration that enforces good practices.
```
sindresorhus: Maybe linkify ``recommended` configuration` to the relative file path of index.js so users can easily see the recommended config.

index.js
wyze: Sure, I just copied what was existing. :)

wyze:
Thanks for the feedback. I went ahead and pushed those changes.

sindresorhus:
Awesome! Thanks so much for helping out :) 🍻

jfmengels:
Nice @wyze!
Quick question: does this have any impact on writing rules?

sindresorhus:
@jfmengels I looked through the [migration guide](http://eslint.org/docs/user-guide/migrating-to-2.0.0) before merging. Doesn't look like it.

jfmengels:
Great, thanks!

wyze:
@jfmengels Not so much on the rules themselves. But I noticed in your PRs you updated the `rulesConfig` property, which is now just under `configs` -> `recommended` -> `rules`. So you would need to rebase and update your PRs to reflect that.

jfmengels:
@wyze Not a big deal, I have a lot to fix and rebase anyway ;)

PR #22, opened 2016-02-27
add no-cb-test rule (fixes #2)

### LINKED ISSUE #2: `no-cb-test` rule
Prevent the use of `test.cb()`. Some projects/companies might want to prevent the use of callback tests in favor of only async functions or returning promises.

jfmengels:
Adds no-cb-test rule (fixes #2) :)

cc @sindresorhus @twada (@twada let us know if you want us to stop pinging you ;) )

### DIFF
--- a/docs/rules/no-cb-test.md
+++ b/docs/rules/no-cb-test.md
@@ -1,0 +1,26 @@
+# Ensure no `test.cb()` is used
+
+Disallow the use of `test.cb()`. We instead recommend using `test()` with an async function or a function returning a promise.
+
+
+## Fail
+
+```js
+import test from 'ava';
+
+test.cb('some test', t => {
+	t.pass();
+	t.end();
+});
+```
+
+
+## Pass
+
+```js
+import test from 'ava';
+
+test('some test', async t => {
+	t.pass();
+});
+```
--- a/readme.md
+++ b/readme.md
@@ -29,6 +29,7 @@ Configure it in `package.json`.
 			"ava"
 		],
 		"rules": {
+			"ava/no-cb-test": 0,
 			"ava/no-identical-title": 2,
 			"ava/no-skip-test": 2,
 			"ava/prefer-power-assert": 0,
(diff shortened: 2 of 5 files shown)


docs/rules/no-cb-test.md
```diff
@@ -0,0 +1,26 @@
+# Ensure no `test.cb()` is used
+
+Disallow the use of `test.cb()`.
```
jfmengels: Couldn't think of a good description, but might be nice to have more explanations.

docs/rules/no-cb-test.md
```diff
@@ -0,0 +1,26 @@
...
+
+
+## Pass
+
+```js
+const test = require('ava');
+
+test('some test', t => {
+	t.pass();
+});
```
sindresorhus:
```js
test('some test', async t => {
	t.pass();
});
```

would be a better example as it's the alternative we recommend over `test.cb()`.

readme.md
```diff
@@ -44,6 +45,7 @@ Configure it in `package.json`.
 
 The rules will only activate in test files.
 
+- [no-cb-test](docs/rules/no-cb-test.md) - Ensure no `test.cb()` is used.
```
sindresorhus: > Ensure `test.cb()` is not used.

?

docs/rules/no-cb-test.md
sindresorhus: > We instead recommend using `test()` with an async function or returning a promise.

?

docs/rules/no-cb-test.md
```diff
@@ -0,0 +1,26 @@
+# Ensure no `test.cb()` is used
+
+Disallow the use of `test.cb()`.
+
+
+## Fail
+
+```js
+const test = require('ava');
```
sindresorhus: Use `import test from 'ava';`

docs/rules/no-cb-test.md
jfmengels: Sounds good! I disabled the rule by default though, as its not a rule that every codebase can afford to enforce.

sindresorhus: Agreed

jfmengels:
@sindresorhus All the recommendations sounded good, updated ;)

sindresorhus:
🎉🍻

jfmengels:
\o/

jfmengels:
First merge not creating conflicts on the others too!

PR #24, opened 2016-02-27
add no-skip-assert rule (fixes #16)

### LINKED ISSUE #16: `no-skip-assert` rule
Rule proposal: Ensure that no assertions are skipped.
Similar to the no-skip-test proposal, but for assertions. They are especially "dangerous" as they do not appear in the end output.

## Fail

```js
const test = require('ava');

test('some title', (t) => {
	t.skip.is(1, 1);
});
```
## Pass

```js
const test = require('ava');

test('some title', (t) => {
	t.is(1, 1);
});
```

I already wrote a rule for this, so I'll submit it if this seems like a good idea to you (after some of the other PRs I proposed get merged)

jfmengels:
add no-skip-assert rule (fixes #16) :)

### DIFF
--- a/docs/rules/no-skip-assert.md
+++ b/docs/rules/no-skip-assert.md
@@ -1,0 +1,25 @@
+# Ensure no assertions are skipped
+
+It's easy to make an assertion skipped with `t.skip.xyz()` and then forget about it.
+
+
+## Fail
+
+```js
+import test from 'ava';
+
+test('some title', t => {
+	t.skip.is(1, 1);
+});
+```
+
+
+## Pass
+
+```js
+import test from 'ava';
+
+test('some title', t => {
+	t.is(1, 1);
+});
+```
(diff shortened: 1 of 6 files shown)


docs/rules/no-skip-assert.md
```diff
@@ -0,0 +1,25 @@
+# Ensure no assertions are skipped
+
+It's easy to make an assertion skipped with `test.skip.xyz()` and then forget about it.
```
sindresorhus: `test.skip.xyz()` => `t.skip.xyz()`

rules/no-skip-assert.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+var createAvaRule = require('../create-ava-rule');
+
+function nameOfRootObject(node) {
```
sindresorhus: Can you put this in `util.js`?

docs/rules/no-skip-assert.md
```diff
@@ -0,0 +1,25 @@
...
+
+
+## Fail
+
+```js
+const test = require('ava');
```
sindresorhus: `import test from 'ava';`

jfmengels:
Updated!

sindresorhus:
:+1: ✨🌸

PR #35, opened 2016-03-04
no-statement-after-end

### LINKED ISSUE #25: `no-statement-after-end` rule
Rule proposal: Enforce that no statements happen after a `t.end()` is called.

Might be a bit hard to handle if `t.end()` is called conditionnally, but it's worth a try.

jamestalmage:
Fixes #25 

reviewers: @jfmengels 

I dropped one assertion from your original test suite. IMO, the following should be legal:

```js
t.end();
return;
```

Specifically so we can accommodate the following:

```js
if (foo) {
  t.end();
  return;
} 

console.log('bar');
t.end();
```


sindresorhus:
Would be helpful with some code comments about the segment stuff. It's a bit hard to follow.

sindresorhus:
// @twada

rules/no-statement-after-end.js
```diff
@@ -0,0 +1,108 @@
...
+		add(segment);
+
+		segment.prevSegments.forEach(add);
+
+		return unreported;
+	}
+
+	function isUnreported(segment) {
+		var info = segmentInfo(segment);
+		return !(!info.endExpression || info.reported);
```
jfmengels: Simplify to `return info.endExpression && !info.reported;`?

rules/no-statement-after-end.js
```diff
@@ -0,0 +1,108 @@
...
+		function add(segment) {
+			if (isUnreported(segment)) {
+				unreported.push(segmentInfo(segment));
+			}
+		}
+		add(segment);
+
+		segment.prevSegments.forEach(add);
+
+		return unreported;
```
jfmengels: I find a functional approach more readable, but that might just be me.
```js
return [segment].concat(segment.prevSegments)
  .filter(isUnReported)
  .map(segmentInfo);
```

jfmengels:
Writing it down just in case: Don't forget to add a doc file and update the readme.

:+1: for `t.end(); return;`

Do you mind adding these 2 tests? They already pass, so no additional work to do for these.
```
header + 'test.cb(function (t) { return t.end(); });',
header + 'test.cb(function (t) { t.end(); return; });'
```

jamestalmage:
> I don't think anything else than an empty return statement should be accepted

[That's exactly what this if statement does](https://github.com/sindresorhus/eslint-plugin-ava/commit/c80d10560f1fa65f3253fbc5b2aa1d55a8267b98#diff-278c010263b3ae6dfe1f8a93ea79d5c8R77)

twada:
> Would be helpful with some code comments about the segment stuff. It's a bit hard to follow.

I agree with @sindresorhus. Yes, I'm AST guy and I can read this pull-req but we need some comments for us.

jamestalmage:
Agreed. I just need to find time to do it! Hopefully this weekend

jamestalmage:
OK,
I added some comments, and refactored a bit. I think it is (somewhat) easier to understand now.

Part of the problem is that `eslint`'s code path analysis utility is just hard to wrap your head around until you play with it for a while. One of the added comments links to that documentation.

jfmengels:
This looks great @jamestalmage. It's still hard to get the head around, but it's much better already. The only missing parts are the doc file for the rule and to update the readme file. I did that already [here][https://github.com/jfmengels/eslint-plugin-ava/commit/f15c22d481a8705555af6c7820c535a97f068238] if you want to reuse it.

jamestalmage:
@jfmengels - Done

readme.md
```diff
@@ -59,6 +59,7 @@ The rules will only activate in test files.
 - [no-only-test](docs/rules/no-only-test.md) - Ensure no `test.only()` are present.
 - [no-skip-assert](docs/rules/no-skip-assert.md) - Ensure no assertions are skipped.
 - [no-skip-test](docs/rules/no-skip-test.md) - Ensure no tests are skipped.
+- [no-statement-after-end](docs/rules/no-statement-after-end.md) - Ensure `t.end()` is the last statement executed.
```
jfmengels: I like the new summary you use here. Could you make that the title of the doc file too?

jamestalmage: done

jfmengels:
LGTM :)

jfmengels:
:sparkles: :smiley:

sindresorhus:
LGTM2

687474703a2f2f7777772e7265616374696f6e676966732e636f6d2f77702d636f6e74656e742f67616c6c6572792f64616e63652d70617274792f74756d626c725f6d306f73727268714d6b31717a6376376e6f345f3235302e676966

PR #36, opened 2016-03-04
config: added ava prefix

solcik:
This pull request is fixing:

```
1:1  error  Definition for rule 'no-identical-title' was not found                                    no-identical-title
1:1  error  Definition for rule 'no-invalid-end' was not found                                        no-invalid-end
1:1  error  Definition for rule 'no-only-test' was not found                                          no-only-test
1:1  error  Definition for rule 'no-skip-assert' was not found                                        no-skip-assert
1:1  error  Definition for rule 'no-skip-test' was not found                                          no-skip-test
1:1  error  Definition for rule 'test-ended' was not found                                            test-ended
1:1  error  Definition for rule 'test-title' was not found                                            test-title
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,17 +24,17 @@ module.exports = {
 				sourceType: 'module'
 			},
 			rules: {
-				'max-asserts': [0, 5],
-				'no-cb-test': 0,
-				'no-identical-title': 2,
-				'no-invalid-end': 2,
-				'no-only-test': 2,
-				'no-skip-assert': 2,
-				'no-skip-test': 2,
-				'no-todo-test': 1,
-				'prefer-power-assert': 0,
-				'test-ended': 2,
-				'test-title': [2, 'always']
+				'ava/max-asserts': [0, 5],
+				'ava/no-cb-test': 0,
+				'ava/no-identical-title': 2,
+				'ava/no-invalid-end': 2,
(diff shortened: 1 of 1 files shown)


sindresorhus:
According to the docs this is not correct: http://eslint.org/docs/developer-guide/working-with-plugins#rules-in-plugins

You're the one that has to prefix rules with `ava/`, not this plugin.

jfmengels:
@sindresorhus The error happens with the recommended settings.
I can replicate the error when using `"extends": "plugin:ava/recommended"` in my eslintrc.

erikdesjardins:
@sindresorhus That's true, but this PR changes the recommended config, not the rules themselves.

See also the recommended configs of [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import/blob/8222cdd5e2e9ecd4536b01eed74984dc44691849/config/errors.js), [eslint-plugin-react](https://github.com/yannickcr/eslint-plugin-react/blob/a3eccd8b30d867660e5fe9506c6ebf6862ff3b6f/index.js#L47-L67).

jfmengels:
I can replicate this, and this fix makes it ok, so this looks good to me. Thanks @solcik!

sindresorhus:
Oh, sorry. I thought I was looking at a different part of that file.

PR #37, opened 2016-03-04
Fix incorrect no-identical-rule errors for title expressions

jfmengels:
I noticed that the rule gets incorrectly triggered when using non-string literals (binary expressions, template literals...). Now I compare the nodes rather than the value of the title (which was undefined in the case of template literals for example...).

I also added a few tests of cases I could think of but that were already ok.

I can still see two edge cases for this rule, but I don't think we can really do much against (thought I should at least write them somewhere):
- Same title node used in different contexts (False error):
```js
let name = 'foo';
test(name, t => {
  t.is(foo(), 1);
});

name = 'bar';
test(name, t => {
  t.is(bar(), 1);
});
```

- Same expression value, but with different nodes (false pass):
```js
test("ab", t => { ... });
test("a"+"b", t => { ... });
```

### DIFF
--- a/rules/no-identical-title.js
+++ b/rules/no-identical-title.js
@@ -1,10 +1,23 @@
 'use strict';
+var espurify = require('espurify');
+var deepStrictEqual = require('deep-strict-equal');
 var createAvaRule = require('../create-ava-rule');
+
+function purify(node) {
+	return node && espurify(node);
+}
+
+function isTitleUsed(usedTitleNodes, titleNode) {
+	var purifiedNode = purify(titleNode);
+	return usedTitleNodes.reduce(function (prev, usedTitle) {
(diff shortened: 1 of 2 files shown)


sindresorhus:
> Same title node used in different contexts (False error):

Should be possible using the ESLint codepath thing.

> Same expression value, but with different nodes (false pass):

Yeah, and such an edge-case it's not really worth wasting energy on.

sindresorhus:
LGTM

jfmengels:
Forgot the mandatory celebration: \o/
:D

sindresorhus:
I know right. Battery was running out so I had to just quickly merge and release.

Better later than never! ------:arrow_down:

d8ee7104-273c-11e5-9676-43241f2fc082

PR #40, opened 2016-03-07
add `use-test` rule (fixes #39)

### LINKED ISSUE #39: `use-test` rule
Rule proposal: Enforce that AVA gets imported with `test` as a variable name.

### Bad
```js
var foo = require('ava');
let foo = require('ava');
const foo = require('ava');
import foo from 'ava';
```

### Good
```js
var test = require('ava');
let test = require('ava');
const test = require('ava');
import test from 'ava';

var test = require('foo');
import test from 'foo';
```

I think this should be a recommended rule at error level, especially considering that all rules assume that AVA is named `test`. What do you think?

jfmengels:
adds `use-test` rule (fixes #39) :)

### DIFF
--- a/docs/rules/use-test.md
+++ b/docs/rules/use-test.md
@@ -1,0 +1,24 @@
+# Ensure that AVA is imported with `test` as the variable name
+
+The convention is to import AVA and assign it to a variable named `test`. Most rules in `eslint-plugin-ava` are based on that assumption.
+
+### Fail
+
+```js
+var ava = require('ava');
+let ava = require('ava');
+const ava = require('ava');
+import ava from 'ava';
+```
+
+### Pass
+
+```js
+var test = require('ava');
+let test = require('ava');
+const test = require('ava');
+import test from 'ava';
+
+var test = require('foo');
+import test from 'foo';
+```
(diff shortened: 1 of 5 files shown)


docs/rules/use-test.md
```diff
@@ -0,0 +1,24 @@
+# Ensure that AVA gets imported with `test` as the variable name
```
sindresorhus: > Ensure that AVA is imported with `test` as the variable name

rules/use-test.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+module.exports = function (context) {
+	return {
+		ImportDeclaration: function (node) {
+			if (node.source.value === 'ava' && node.specifiers[0].local.name !== 'test') {
+				report(context, node);
+			}
+		},
+
```
sindresorhus: No empty line here

rules/use-test.js
```diff
@@ -0,0 +1,37 @@
...
+	arguments: [
+		{
+			type: 'Literal',
+			value: 'ava'
+		}
+	]
+};
+
+function report(context, node) {
+	context.report(node, 'AVA should be imported as test.');
```
sindresorhus:
```js
context.report(node, 'AVA should be imported as `test`.');
```

jfmengels:
:+1:. Updated

sindresorhus:
Woot! :dancer: 

tumblr_mqmk6j8kkz1qdlh1io1_250

PR #41, opened 2016-03-07
add `use-t` rule (fixes #38)

### LINKED ISSUE #38: `use-t` rule
Rule proposal: Enforce that the test function has either no parameters, or only one parameter named `t`.

### Bad
```js
test(foo => { // Incorrect name
  // ...
});

test((foo) => { // Incorrect name
  // ...
});

test((t, bar) => { // too many arguments
  // ...
});

test((bar, t) => { // too many arguments
  // ...
});
```

### Good
```js
test(t => {
  // ...
});

test(foo); // As I don't think we can infer the arguments of foo, or even if it is a function

test.todo("foo");
```

I think this should be a recommended rule at error level, especially considering that all rules assume that `t` is indeed called `t`. What do you think?

jfmengels:
adds `use-t` rule (fixes #38) :)

### DIFF
--- a/docs/rules/use-t.md
+++ b/docs/rules/use-t.md
@@ -1,0 +1,35 @@
+# Ensure test functions use `t` as their parameter
+
+The convention is to have the parameter in AVA's test function be named `t`. Most rules in `eslint-plugin-ava` are based on that assumption.
+
+### Fail
+
+```js
+import test from 'ava';
+
+test(foo => { // Incorrect name
+	t.pass();
+});
+
+test((t, bar) => { // too many arguments
+	t.pass();
+});
+
+test((bar, t) => { // too many arguments
+	t.pass();
+});
+```
+
+### Pass
+
+```js
+import test from 'ava';
+
+test(() => {
+	// ...
+});
+
+test(t => {
+	t.pass();
+});
+```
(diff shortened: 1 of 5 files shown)


docs/rules/use-t.md
```diff
@@ -0,0 +1,39 @@
+# Ensure test functions use `t` as their parameter
+
+The convention is to have the parameter in AVA's test function be named `t`. Most rules in `eslint-plugin-ava` are based on the assumption.
```
sindresorhus: `the` => `that`

docs/rules/use-t.md
```diff
@@ -0,0 +1,39 @@
...
+### Fail
+
+```js
+import test from 'ava';
+
+test(foo => { // Incorrect name
+	t.pass();
+});
+
+test((foo) => { // Incorrect name
```
sindresorhus: This is the same as the above, so I don't see the point in showing that.

jfmengels:
:+1:. Updated

sindresorhus:

jfmengels:
Done!

sindresorhus:
tumblr_mpamnkvlcv1qdlh1io1_400

PR #42, opened 2016-03-12
Fix errors with `test.todo()`

bdougherty:
Fixes a couple issues when using `test.todo()`:

* `test-title` incorrectly reports that the test doesn’t have a title when it does
* `no-identical-title` wasn’t pulling the right title, so if you had multiple `test.todo()`s in the file, it would error when it shouldn’t have

### DIFF
--- a/rules/no-identical-title.js
+++ b/rules/no-identical-title.js
@@ -26,7 +26,7 @@ module.exports = function (context) {
 			}
 
 			var args = node.arguments;
-			var titleNode = args.length > 1 ? args[0] : undefined;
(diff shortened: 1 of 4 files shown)


sindresorhus:
Looks good to me.

@jfmengels ?

jfmengels:
Looks good, thanks a lot @bdougherty!
This also pretty much saves us the trouble to write a 'todo-needs-title' rule.

sindresorhus:
Cool. Thanks @bdougherty :)

PR #44, opened 2016-03-20
Fix `no-identical-title` for dynamic template strings

### LINKED ISSUE #43: no-identical-title and template strings
I'm just tying out a test like this:

```js
;[
  ['BIT_STRING', BIT_STRING],
  ['CHARACTER_STRING', CHARACTER_STRING],
  ['OCTET_STRING', OCTET_STRING]
].forEach(([title, type]) => {
  test(`decodes ${title}`, t => {
    const expected = new Buffer('foo')
    t.same(decode(type, 3, expected, 0), expected)
  })
})

;[
  ['ENUMERATED', ENUMERATED],
  ['INTEGER', INTEGER]
].forEach(([title, type]) => {
  test(`decodes ${title}`, t => {
    t.is(decode(type, 1, new Buffer([42]), 0), 42)
    const expected = new Buffer(7)
    t.same(decode(type, 7, expected, 0), expected)
  })
})
```

The `no-identical-title` complains about a duplicate title in `decodes ${title}`. It should probably recognize the title is a template string and is most likely not a duplicate.

jfmengels:
Fixes #43.
The `no-indentical-rule` now tries to determine whether the title is dynamic (= contains a variable) or not, so that it doesn't flag titles like `\ ̀ ensure x=${expected} \ ̀`, but still flags titles like `\ ̀ ensure x=2 \ ̀` or `\ ̀ ensure x=${1 + 3}\ ̀` (template strings without any variables). Did the same thing for titles like "foo" + 1.
I feel like I'm forgetting cases, but it should fix #43 at least.

Note: also simplified the tests to use `t => {}` rather than `function() {}`

### DIFF
--- a/rules/no-identical-title.js
+++ b/rules/no-identical-title.js
@@ -7,11 +7,22 @@ function purify(node) {
 	return node && espurify(node);
 }
 
+function isStaticTemplateLiteral(node) {
+	return node.expressions.every(isStatic);
+}
+
+function isStatic(node) {
+	return !node ||
+		node.type === 'Literal' ||
+		(node.type === 'TemplateLiteral' && isStaticTemplateLiteral(node)) ||
+		(node.type === 'BinaryExpression' && isStatic(node.left) && isStatic(node.right));
+}
+
 function isTitleUsed(usedTitleNodes, titleNode) {
 	var purifiedNode = purify(titleNode);
-	return usedTitleNodes.reduce(function (prev, usedTitle) {
-		return prev || deepStrictEqual(purifiedNode, usedTitle);
-	}, false);
+	return usedTitleNodes.some(function (usedTitle) {
+		return deepStrictEqual(purifiedNode, usedTitle);
+	});
 }
 
 /* eslint quote-props: [2, "as-needed"] */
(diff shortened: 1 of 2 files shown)


novemberborn:
Nice, thanks!

> Note: also simplified the tests to use t => {} rather than function() {}

For next time, it'd be better to do that in its own commit. Right now it makes it harder to see the actual changes.

jfmengels:
@novemberborn you're right, my bad. If anyone wants it, I can go back and revert those changes and do a pass afterwards.

rules/no-identical-title.js
```diff
@@ -7,6 +7,19 @@ function purify(node) {
 	return node && espurify(node);
 }
 
+function isStaticTemplateLiteral(node) {
+	return node.expressions.reduce(function (res, expr) {
+		return res && isStatic(expr);
+	}, true);
```
twada:
```js
return node.expressions.every(isStatic);
```
would be better.

jfmengels: Man, didn't even know `every` and `some` were built-in array methods. You learn everyday.

sindresorhus:
Thanks for fixing @jfmengels :+1: 🎉

jfmengels:
No problem :)

PR #53, opened 2016-03-24
Do not enforce a title on hooks (fixes #52)

### LINKED ISSUE #52: test.title needed for test.before() and friends
image
I've been told that not giving a title to a `t.before()` (and friends) triggers the ̀test-title` rule, and I wondered what you guys thought about this. Should we disable this by default for those, and allow them to be enabled with an option?
I don't really see the point of having a title for a before/after, but some might find it useful. Anyway, I think it should be disabled by default at the very least.

jfmengels:
Do not enforce a title on hooks (fixes #52)


sindresorhus:
3kpq4m

Issue #54, opened 2016-03-24
`no-incompatible-modifiers` rule


jfmengels:
Proposal: Create a rule that disallows certain combination of modifiers, such as
```js
test.only.skip(...)
test.before.only(...)
test.before.skip(...) // ?
test.before.beforeEach()
test.before.after()
test.before.serial()
// others?
```

Value: Earlier feedback when linting in the editor

Let me know if you think that this is an unnecessary rule or not (it probably is, or very close to).

jamestalmage:
I would say unnecessary, same as https://github.com/sindresorhus/eslint-plugin-ava/pull/55#issuecomment-201120065. It should be a runtime error, not a linter one.

People are not likely to make these mistakes often (they make no sense), so there isn't a lot of value.

jfmengels:
Agree these are pretty rare cases, just wanted to have your thoughts on this ;)

PR #55, opened 2016-03-25
`no-unknown-modifiers` rule

jfmengels:
Adds `no-unknown-modifiers` rule

Detects unknown `test` modifiers
```js
test.onlu(...)
test.ski(...)
test.Before(...)
test.beforeeach(...)
```

I have omitted 'default' and 'test' from the modifier list, as I think those are simply import noise. Let me know what you think.


jamestalmage:
Hmm. I guess I don't get the value. These types of mistakes are going to surface at runtime. Why do all this work to detect a problem that will be detected anyways?

jfmengels:
This rule is only useful when you have a linter in your editor, and it's not the only rule that has been written for that purpose (like: `no-invalid-end`).
I agree that #54 is pretty useless as it pretty much never happens, but catching this kind of common error (typos...) very early is definitely useful IMO. Outside of an editor, I agree that it has no value.

sindresorhus:
I think it would be better to just hardcode the list of modifiers. It's not going to get updated very often. Hopefully never after 1.0.0. And we control both so we can easily update. Requiring `ava` comes with unnecessary overhead.

sindresorhus:
I first thought what James said, but after thinking about it, it's handy for catching early errors.

jamestalmage:
:+1:

jfmengels:
Ok, I will hardcode the modifiers then. I'll probably make a similar rule for assertions later on.

jfmengels:
Updated the PR and its description. It's not a WIP anymore.

rules/no-unknown-modifiers.js
```diff
@@ -0,0 +1,46 @@
...
+	after: true,
+	afterEach: true,
+	before: true,
+	beforeEach: true,
+	cb: true,
+	only: true,
+	serial: true,
+	skip: true,
+	todo: true
+};
```
sindresorhus: Why not just an array? You won't gain any noticeable speed improvement for such a small list.

rules/no-unknown-modifiers.js
```diff
@@ -0,0 +1,46 @@
...
+
+function getTestModifiers(node) {
+	if (node.type === 'CallExpression') {
+		return getTestModifiers(node.callee);
+	}
```
sindresorhus: empty line below all blocks

rules/no-unknown-modifiers.js
```diff
@@ -0,0 +1,46 @@
...
+	if (node.type === 'MemberExpression') {
+		return getTestModifiers(node.object).concat(node.property.name);
+	}
```
sindresorhus: empty line

rules/no-unknown-modifiers.js
```diff
@@ -0,0 +1,46 @@
...
+
+function unknownModifiers(node) {
+	return getTestModifiers(node)
+		.filter(function (modifier) {
+			return !modifiers[modifier];
+		});
+}
+
+module.exports = function (context) {
+	var ava = createAvaRule();
```
sindresorhus: empty line below here

rules/no-unknown-modifiers.js
```diff
@@ -0,0 +1,46 @@
...
+	return ava.merge({
+		CallExpression: function (node) {
+			if (!ava.isTestFile || ava.currentTestNode !== node) {
+				return;
+			}
```
sindresorhus: empy line

rules/no-unknown-modifiers.js
```diff
@@ -0,0 +1,46 @@
...
+			var unknown = unknownModifiers(node);
```
sindresorhus: empty line

rules/no-unknown-modifiers.js
```diff
@@ -0,0 +1,46 @@
...
+			if (unknown.length !== 0) {
+				context.report(node, 'Unknown modifier "' + unknown[0] + '"');
```
sindresorhus: Use 

```
`
```

instead of `"`

test/no-unknown-modifiers.js
```diff
@@ -0,0 +1,82 @@
...
+			`${header} test.skip(t => {});`,
+			`${header} test.todo(t => {});`,
+			// shouldn't be triggered since it's not a test file
+			`test.foo(t => {});`
+		],
+		invalid: [
+			{
+				code: `${header} test.foo(t => {});`,
+				errors: [
+          {...ruleError, message: 'Unknown modifier "foo"'}
```
sindresorhus: inconsidtent indent

rules/no-unknown-modifiers.js
jfmengels: Was not a performance related decision. I found `modifiers[x]` was just a pretty way to know whether x is in there. But yeah, it's more verbose in the list declaration bit, and a tad more error-prone (`{ ..., value: false, ...`). Changed.

rules/no-unknown-modifiers.js
jfmengels: Sure thing, done ;)
I'd say make it a xo rule, but it's not a rule that I like much, and I'm starting to use xo more and more, so I won't say anything ;)

sindresorhus:
LGTM

jfmengels:
Updated

sindresorhus:
Thanks @jfmengels :)

Internet dropped so couldn't post a gif yesterday.

leo-excited

PR #56, opened 2016-03-30
Add no-ignored-test-files rule (fixes #51)

### LINKED ISSUE #51: Check parent directory name
In a rule check that we do not create a test files (a file containing `import test from 'ava';`?) in a path that is not in an ignored path like `fixtures` or `helpers`. 

- `test/units/foo/myTestFile.js` should not trigger an error 
- `test/units/foo/helpers/myHelper.js` should not trigger an error 
- `test/units/foo/helpers/myTestFile.js` should trigger an error

jfmengels:
Add no-ignored-test-files rule (fixes #51)

I would like to add more checks to see whether the test is in the AVA glob (with an option to de-/activate or set a custom glob), but I haven't found (yet, haven't searched much for now) a library that simply tells me whether a file matches a glob, and I don't know if we can reliably get the cwd from ESLint (or the editor running ESLint).
This can be added later to the rule though. I might take a look at it later, or someone could take a look too if interested.


novemberborn:
> I would like to add more checks to see whether the test is in the AVA glob (with an option to de-/activate or set a custom glob), but I haven't found (yet, haven't searched much for now) a library that simply tells me whether a file matches a glob

The watcher does this, see <https://github.com/sindresorhus/ava/blob/005b656b9950ab3fa1e2bd9af326d5b39540e71b/lib/watcher.js#L333:L379>. It uses <https://www.npmjs.com/package/multimatch>.

jfmengels:
From what I could test, in Atom's ESLint plugin (the Atom plugin that runs ESLint I mean), `process.cwd()` returns the current file's folder everytime, not the root of the project, and ESLint's `context.getFilename()` returns the file's absolute path. Meaning it becomes pretty hard to figure out the project's root folder.

So it means two things:
- Globbing becomes pretty hard, because even the default glob is relative to the folder Ava is run from.
- Anybody who uses eslint-plugin-ava (at least in Atom) will have this rule ringing loudly for every file if his project, dev folder, or any higher-up folder is named fixtures/helpers.

The second is obviously pretty rare (I'd think at least), but it can be problematic for some users. Let me know what you guys think about this.

novemberborn:
@jfmengels could use https://www.npmjs.com/package/pkg-up until you find a package that depends on AVA?

jfmengels:
@novemberborn Sounded good, so I tried it. I disabled the rule altogether progammatically when no `package.json` is found (debatable, but probably pretty rare anyway).

The result is pretty rad IMO!
screenshot from 2016-04-02 20-08-10

rules/no-ignored-test-files.js
```diff
@@ -0,0 +1,92 @@
...
+function isInIgnoredFolder(relativeFilePath) {
+	var folders = relativeFilePath.split(path.sep);
+	var ignoredFolder;
+	if (folders.indexOf('fixtures') !== -1) {
+		ignoredFolder = 'fixtures';
+	} else if (folders.indexOf('helpers') !== -1) {
+		ignoredFolder = 'helpers';
+	}
+
+	if (ignoredFolder) {
```
novemberborn: Why not use `multimatch` here as well? Copy [`excludePatterns`](https://github.com/sindresorhus/ava/blob/005b656b9950ab3fa1e2bd9af326d5b39540e71b/api.js#L34:L37) without the leading `!` and you should be good to go.

rules/no-ignored-test-files.js
```diff
@@ -0,0 +1,92 @@
...
+}
+
+function isIgnored(rootDir, files, filepath) {
+	var relativeFilePath = path.relative(rootDir, filepath);
+	return isInIgnoredFolder(relativeFilePath) ||
+		isNotInGlob(files, relativeFilePath);
+}
+
+function getPackageInfo() {
+	var packageFilePath = pkgUp.sync();
```
novemberborn: Might be good to document the assumption that `process.cwd()` is the directory of the file being linted?

Some test suites may have nested `package.json` files. I wonder if you should keep searching until you find AVA config?

[Here's an example from AVA itself with a nested `package.json`](https://github.com/sindresorhus/ava/tree/005b656b9950ab3fa1e2bd9af326d5b39540e71b/test/fixture/caching).

novemberborn:
Do people put test factories in their helpers? That might be a legitimate `ava` import. Though you could disable the rule in that case or customize the pattern.

---

Very cool @jfmengels!

rules/no-ignored-test-files.js
jfmengels: Sounds good! Will do that (won't put node_modules though as ESLint ignores them already)

rules/no-ignored-test-files.js
jfmengels: I'll change the wording to indicate that it will use the closest package.json. I think in most cases having multiple `package.json` will be used for examples inside the repo or test fixtures (though not sure I see the value in the one you showed).

> I wonder if you should keep searching until you find AVA config?

Why not, but then again, it will not be able to handle some cases. Some projects might have sub-projects as examples (thinking of [Redux](https://github.com/reactjs/redux/tree/master/examples, even though it doesn't use AVA afaik) where it's pretty much the point that they don't depend on any parent dirs. I'm guessing that in most cases, what you want is isolation from the parent "project", so I'd rather not do this. But let me know what you guys think.

jfmengels:
@novemberborn @jamestalmage had the same comment https://github.com/sindresorhus/eslint-plugin-ava/issues/51#issuecomment-201146646, and I agree that you should then simply disable the rule. There was an interesting proposition https://github.com/sindresorhus/ava/issues/695 for test macros which would be better. We only lint an error if you import AVA and create a test, but we could then make an exception if you're defining a macro

jfmengels:
Updated :)

novemberborn:
> We only lint an error if you import AVA and create a test, but we could then make an exception if you're defining a macro

That sounds good.

docs/rules/no-ignored-test-files.md
```diff
@@ -0,0 +1,92 @@
+# Ensure no tests are written in ignored files
+
+When searching for tests, AVA ignores files contained in folders named `fixtures` or `helpers`. By default, it will search in `test.js test-*.js test/**/*.js`, which you can override by specifying a path when launching AVA or in the [AVA configuration in the `package.json` file] ](https://github.com/sindresorhus/ava#configuration).
```
novemberborn: Mention `node_modules` here?

novemberborn:
Nothing further from me :dancers:

docs/rules/no-ignored-test-files.md
jfmengels: Well, ESLint ignores `node_modules` anyway. I don't really mind adding it, but for consistency we'd have to add it to every rule. I think that users of Node or ESLint know that node_modules/ should not be touched.

novemberborn: Yes but you're describing AVA's behavior here.

jfmengels: True. I've added the mention about node_modules, and to clarify, a message sayig that they won't be linted due to ESLint.

sindresorhus:
LGTM

jfmengels:
Cool, I'll squash and merge then.

jamestalmage:
@jfmengels - FYI - GH has a new "Squash and Merge" feature built in. Just click the merge button.

Ideally you would use that, even for 1 commit PR's. It avoids creating an extra merge commit.

If there are multiple commits that you want to keep separate, the old merge method is fine.

jfmengels:
Aah. Yeah I knew about it, but I thought it was something that you needed to activate as I didn't see it. But it only shows when you click the merge button once apparently ^^'

(Hoping for a rebase button to come soon)

jamestalmage:
You usually only need to rebase if there is a merge conflict (in which case you probably want to do some edits / testing locally). I guess it could be useful for selectively squashing (i.e. ending up with less commits, but still more than one), and for editing commit messages. But that is a rare enough need I'm not clamoring for it. I would much rather see improvements to issue management first.

jfmengels:
I mean rebase in the sense of no merge commits. I like straight lines.

sindresorhus:
> I mean merge rebase in the sense of no merge commits. I like straight lines.

That's exactly what the squash button does, but you seems you used the old merge button.

PR #65, opened 2016-04-06
fix(test-ended): false positive when passing `t.end` as a callback. Fixes #60.

### LINKED ISSUE #60: False positive in the `test-ended` rule when `t.end` is passed as an argument
The following results in a violation, but it's a common pattern, so I think we should support it.

```js
import test from 'ava';

test.cb(t => {
	setTimeout(t.end, 10);
});
```

jamestalmage:
Fixes #60.

### DIFF
--- a/docs/rules/test-ended.md
+++ b/docs/rules/test-ended.md
@@ -23,4 +23,8 @@ test.cb(t => {
 	t.pass();
 	t.end();
 });
+
+test.cb(t => {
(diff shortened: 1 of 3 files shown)


sindresorhus:
LGTM

jfmengels:
LGTM

sindresorhus:
Woot! :)

PR #67, opened 2016-04-06
Use off/warn/error in rule configurations

jfmengels:
Changing all rule configurations from using 0/1/2 to using off/warn/error in rule configurations.
This would be consistent with ESLint's own [core config](https://github.com/eslint/eslint/blob/master/conf/eslint.json), and is easier to understood for beginners.

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,22 +29,22 @@ module.exports = {
 				sourceType: 'module'
 			},
 			rules: {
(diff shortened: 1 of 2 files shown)


sindresorhus:
Good ide! This is so much clearer. Especially `'ava/max-asserts': [0, 5],` was super confusing.

PR #68, opened 2016-04-06
Change `test-title` default option (fixes #61)

### LINKED ISSUE #61: Change the default option of the `test-title` rule
Currently the default is `always`, but I honestly think it would be better to have `if-multiple` as a default.

https://github.com/sindresorhus/eslint-plugin-ava/blob/master/docs/rules/test-title.md

There's really no point in forcing naming a test if there's only one, and if users really want this they can just change the default.

Thoughts?

jfmengels:
Changes `test-title` default option (fixes #61)

Did a lof of changes to the tests (sorry...), but didn't update the tests cases themselves much.

### DIFF
--- a/docs/rules/test-title.md
+++ b/docs/rules/test-title.md
@@ -6,6 +6,18 @@ Tests should have a title.
 ## Fail
 
 ```js
+/*eslint ava/test-title: ["error", "if-multiple"]*/
+import test from 'ava';
+
+test(t => {
+	t.pass();
+});
+
+test(t => {
+	t.pass();
+});
+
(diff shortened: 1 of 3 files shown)


sindresorhus:
Thanks for the quick fix! I went with the 10 second rule. 2.0.1 published (naughty I know).

jfmengels:
No pb ;)
And yes it's naughty, though I agree. I've entered OSS by working on Lodash, and there's been a lot of things that could have been considered as breaking changes in v4 (especially wrt FP), so I'm used to it now :D.

(besides, this is debatably a breaking change, since we enforce less, not more)

PR #78, opened 2016-04-11
Add `use-true-false` rule (fixes #77)

### LINKED ISSUE #77: Detect boolean assertions, mandate `t.true()`
Assertions like this evaluate to `true` or `false`:

```js
t.truthy(array.indexOf(value) !== -1);
```

We should require you to write them like:

```js
t.true(array.indexOf(value) !== -1);
```

jfmengels:
Adds `use-true` rule (fixes #77)

### DIFF
--- a/docs/rules/use-true-false.md
+++ b/docs/rules/use-true-false.md
@@ -1,0 +1,33 @@
+# Ensure that `t.true()`/`t.false()` are used instead of `t.truthy()`/`t.falsy()`
+
+`t.true()` and `t.false()` are stricter in their checks than `t.truthy()` and `t.falsy`.
+For example: if you have a function `foo()` which normally returns `true`, but suddenly returns `1` instead, `t.truthy(foo())` would not catch the change, but `t.true(foo())` would.
+This rule enforces the use of the former when the tested expression is known to result in a boolean value.
+
+### Fail
+
+```js
+import ava from 'ava';
+
+test(t => {
+	t.truthy(value < 2);
+	t.truthy(value === 1);
+	t.falsy([1, 2, 3].indexOf(value) === -1);
+	t.falsy(!value);
+	t.truthy(!!value);
+});
+```
+
+### Pass
+
+```js
+import ava from 'ava';
+
+test(t => {
+	t.true(value < 2);
+	t.true(value === 1);
+	t.false([1, 2, 3].indexOf(value) === -1);
+	t.false(!value);
+	t.true(!!value);
+});
+```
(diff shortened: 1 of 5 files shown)


sindresorhus:
The name is a bit misleading as it supports `t.false` too.

Some other suggestions:

- `use-true-false`
- `use-strict-equal`
- `use-strict-equality`

jfmengels:
Yeah, I changed it to `use-true-false`

docs/rules/use-true-false.md
```diff
@@ -0,0 +1,31 @@
+# Ensure that `t.true()`/`t.false()` are used instead of `t.truthy()`/`t.falsy()`
+
+`t.true()` and `t.false()` are stricter in their checks than `t.truthy()` and `t.falsy`. This rule enforces the use of the former when the tested expression is known to result in a boolean value.
```
forresst: For a better understanding, I would add the example of @sindresorhus:

```markdown
`t.true()` and `t.false()` are stricter in their checks than `t.truthy()` and `t.falsy`.
For example: imagine you have a function foo() which normally returns true, but suddenly returns 1 instead. t.truthy(foo()) would not catch that case.
This rule enforces the use of the former when the tested expression is known to result in a boolean value.
```

docs/rules/use-true-false.md
```diff
@@ -0,0 +1,33 @@
+# Ensure that `t.true()`/`t.false()` are used instead of `t.truthy()`/`t.falsy()`
+
+`t.true()` and `t.false()` are stricter in their checks than `t.truthy()` and `t.falsy`.
+For example: if you have a function `foo()` which normally returns true, but suddenly returns 1 instead, `t.truthy(foo())` would not catch the change, but `t.true(foo())` would.
```
novemberborn: Maybe use backticks around "true" and "1"?

novemberborn:
Very cool @jfmengels!

sindresorhus:
Maybe also support builtin methods known to return booleans? I'm thinking `Array.isArray()`, etc.

jfmengels:
@sindresorhus Would be nice yeah. Do you have a package that lists them by any chance?

sindresorhus:
No

sindresorhus:
Landed this. We can add improvements later: https://github.com/sindresorhus/eslint-plugin-ava/issues/79

Thanks @jfmengels! You're amazingly productive at this.

leo-clapping

jfmengels:
You're welcome ;)

Issue #79, opened 2016-04-12
Add more supported cases for the `use-true-false` rule


sindresorhus:
See: https://github.com/sindresorhus/eslint-plugin-ava/pull/78#issuecomment-208840730

- [ ] We'll need to compile a list of builtin methods that return booleans. Help wanted for that.

PR #82, opened 2016-04-13
use-true-false: Add framework for supporting known boolean expressions.

jamestalmage:
Partial Fix: #79

I didn't hunt down a complete list of boolean expressions, but I have provided a framework that makes it easy to edit the list.

It uses `call-signature` (which is part of what `power-assert` uses to do some of it's magic).

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
   "dependencies": {
     "call-signature": "0.0.2",
     "deep-strict-equal": "^0.1.0",
+    "espree": "^3.1.3",
     "espurify": "^1.5.0",
     "multimatch": "^2.1.0",
     "object-assign": "^4.0.1",
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "mocha"
   ],
   "dependencies": {
+    "call-signature": "0.0.2",
     "deep-strict-equal": "^0.1.0",
     "espurify": "^1.5.0",
     "multimatch": "^2.1.0",
--- a/package.json
+++ b/package.json
@@ -61,7 +61,6 @@
     "mocha"
   ],
   "dependencies": {
-    "call-signature": "0.0.2",
     "deep-strict-equal": "^0.1.0",
     "espree": "^3.1.3",
     "espurify": "^1.5.0",
--- a/rules/use-true-false.js
+++ b/rules/use-true-false.js
@@ -1,7 +1,9 @@
 'use strict';
+var espree = require('espree');
+var espurify = require('espurify');
+var deepStrictEqual = require('deep-strict-equal');
 var util = require('../util');
 var createAvaRule = require('../create-ava-rule');
-var callSignature = require('call-signature');
 
 var booleanBinaryOperators = [
 	'==', '===', '!=', '!==', '<', '<=', '>', '>='
@@ -10,7 +12,7 @@ var booleanBinaryOperators = [
 var knownBooleanSignatures = [
 	'Array.isArray(x)'
 ].map(function (signature) {
-	return callSignature.parse(signature).callee;
+	return espurify(espree.parse(signature).body[0].expression.callee);
 });
 
 module.exports = function (context) {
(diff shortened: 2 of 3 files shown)


sindresorhus:
Awesome. My contribution to this, I took the time to compile a list of builtins returning a boolean:

```
isFinite()
isNaN()
Object.is()
Object.isExtensible()
Object.isFrozen()
Object.isSealed()
Boolean()
Number.isNaN()
Number.isFinite()
Number.isInteger()
Number.isSafeInteger()
String.prototype.includes()
String.prototype.endsWith()
String.prototype.startsWith()
RegExp.prototype.test()
Array.isArray()
Array.prototype.includes()
Array.prototype.every()
Array.prototype.some()
Int8Array.prototype.every()
Int8Array.prototype.includes()
Int8Array.prototype.some()
Uint8Array.prototype.every()
Uint8Array.prototype.includes()
Uint8Array.prototype.some()
Uint8ClampedArray.prototype.every()
Uint8ClampedArray.prototype.includes()
Uint8ClampedArray.prototype.some()
Int16Array.prototype.every()
Int16Array.prototype.includes()
Int16Array.prototype.some()
Uint16Array.prototype.every()
Uint16Array.prototype.includes()
Uint16Array.prototype.some()
Int32Array.prototype.every()
Int32Array.prototype.includes()
Int32Array.prototype.some()
Uint32Array.prototype.every()
Uint32Array.prototype.includes()
Uint32Array.prototype.some()
Float32Array.prototype.every()
Float32Array.prototype.includes()
Float32Array.prototype.some()
Float64Array.prototype.every()
Float64Array.prototype.includes()
Float64Array.prototype.some()
Map.prototype.has()
Set.prototype.has()
WeakMap.prototype.has()
WeakSet.prototype.has()
ArrayBuffer.isView()
SharedArrayBuffer.isView()
Reflect.has()
Reflect.isExtensible()
```

Should I put in a module? Would there be any value to that?

jamestalmage:
> should I put it in a module? Would there be any value to that?

Maybe - who knows. Probably more so if it had the full method signature:

```
Array.isArray(obj)
Object.is(value1, value2)
obj.fn(arg1, arg2, [optional_arg])
```

Also, `call-signature` is not smart enough to parse `String.prototype.includes`. It only allows member expressions one deep. It was a simplification of [`escallmatch`](https://github.com/twada/escallmatch) that uses `regexp` to parse the signature to avoid depending on a full JS parser (`escallmatch` relies on esprima). We could just use `espree` to parse the matcher templates, that would give more robust parsing without adding another dependency.

Still, I am not sure how to handle stuff on the prototype. Were you hoping to try to implement static analysis and figure out if `x.some()` was being called on an Array?

twada:
> We could just use espree to parse the matcher templates, that would give more robust parsing without adding another dependency.

:+1: on @jamestalmage

jamestalmage:
updated to use espree. I didn't realize we already had `espurify` and `deep-strict-equal` as dependencies. Those make the matcher really easy.

I haven't added the full list of methods yet. Still waiting on an answer to this:

> Were you hoping to try to implement static analysis and figure out if x.some() was being called on an Array?

Question was directed at @sindresorhus - but feel free to chime in.

package.json
```diff
@@ -61,7 +61,9 @@
     "mocha"
   ],
   "dependencies": {
+    "call-signature": "0.0.2",
```
jfmengels: Not needed anymore

jfmengels:
This looks good! Nice work @jamestalmage!

jamestalmage:
I went with just the non `prototype` methods, as doing the rest well would require an enormous amount of static type analysis/inferencing.

It would be cool if someone built a linter with a type analysis engine, but I think that's a bit too ambitious to achieve this one feature.

sindresorhus:
> I went with just the non prototype methods, as doing the rest well would require an enormous amount of static type analysis/inferencing.

:+1: 

> It would be cool if someone built a linter with a type analysis engine, but I think that's a bit too ambitious to achieve this one feature.

That is my dream. So many of my rule suggestions on ESLint has been turned down for requiring type analysis... Linting could be so much better with best effort type inference. [Flow](https://github.com/facebook/flow) compiled down to asm.js anyone??

sindresorhus:
Can you add an example of `Array.isArray()` to its docs?

sindresorhus:
Looks good to me. @twada @jfmengels Can you give it a final review?

jfmengels:
LGTM :)

jamestalmage:
> Flow compiled down to asm.js anyone

Even without Flow, Webstorm does an impressive job of type inference from pure JS code. I am sure that was no small undertaking on IntelliJ's part, but I think there are a number of portential uses for an open source alternative. (Atom would surely benefit)

sindresorhus:
@jamestalmage Flow does that too. Its type notation is optional and only there to improve the inference.

There's also https://github.com/ternjs/tern

twada:
LGTM :+1:

sindresorhus:
Awesome! This makes the rule really solid.

jamestalmage:
> Flow does that too. Its type notation is optional and only there to improve the inference

I guess I knew that. I haven't played with Flow much. Does it automatically analyze exports of imported files? It looks line `tern` does, though it appears to require launching a server.

PR #85, opened 2016-04-16
no-commented-tests

### LINKED ISSUE #84: Prevent commented out tests
We should recommend using `test.skip()` instead.

This:

``` js
// test('foo', => {
//  t.pass();
// });
```

Should instead be:

``` js
test.skip('foo', => {
    t.pass();
});
```

jamestalmage:
Prevents commented out tests.

Fixes #84.

Still needs docs.

What should the rule name be? Some ideas: 
- `no-commented-tests`
- `no-commented-out-tests`
- `no-tests-in-comments`
- `no-comment-tests`

### DIFF
--- a/rules/no-commented-test.js
+++ b/rules/no-commented-test.js
@@ -1,5 +1,6 @@
 'use strict';
-var acorn = require('acorn/dist/acorn_loose');
+var acorn = require('acorn');
+var acornLoose = require('acorn/dist/acorn_loose');
 var estraverse = require('estraverse');
 var numLines = require('num-lines');
 var createAvaRule = require('../create-ava-rule');
@@ -31,22 +32,23 @@ module.exports = function (context) {
 		var src = concatComments(commentBlock);
 
 		// Does a best effort parse of the code. AST will contain `x` Identifiers for chunks it can't parse.
-		var ast = acorn.parse_dammit(src, {locations: true});
+		var ast = acornLoose.parse_dammit(src, {locations: true});
 
 		// Traverse the contents of the comment block, looking for a test call.
 		estraverse.traverse(ast, {
 			enter: function (node) {
-				if (node.type !== 'CallExpression') {
-					return;
-				}
+				if (node.type === 'CallExpression' && isTestFunctionCall(node.callee)) {
+					try {
+						// Strict parse - don't report if the commented out test contains syntax errors
+						acorn.parse(src.substring(node.start, node.end));
 
-				if (isTestFunctionCall(node.callee)) {
-					var loc = translateLocation(node.loc.start, commentBlock);
+						var loc = translateLocation(node.loc.start, commentBlock);
 
-					context.report({
-						message: 'use test.skip instead of commenting out tests',
-						loc: loc
-					});
+						context.report({
+							message: 'use test.skip instead of commenting out tests',
+							loc: loc
+						});
+					} catch (e) {}
 				}
 			}
 		});
--- a/rules/no-commented-test.js
+++ b/rules/no-commented-test.js
@@ -1,0 +1,110 @@
+'use strict';
+var acorn = require('acorn/dist/acorn_loose');
+var estraverse = require('estraverse');
+var numLines = require('num-lines');
+var createAvaRule = require('../create-ava-rule');
+
+module.exports = function (context) {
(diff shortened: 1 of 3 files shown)


jamestalmage:
Looks like I am [misusing `parse_dammit`](https://github.com/ternjs/acorn/issues/199#issuecomment-69906491).

jamestalmage:
Some more ideas for this:

## Speed

If I had to guess, parsing every comment is going to incur some significant performance penalties. Especially since I will have to parse most comments twice (see `parse_dammit` link above). It will likely be faster to first scan the comment block with a regular expression:

```js
/test(?:\(|\.)/.test(src)
```

## Accuracy

Should we maybe require valid JS between the opening `(` and ending `)` of `test(...)`?
I am thinking someone might do this:

```js
// can't use test(...) here, because
test.cb(t => { });
```
 
Or they might be cutting / pasting code out of the comment to a refactored test (leaving it in a non-parseable state).

## Disabling

Sometimes your comments contain example code. Especially if you are using JSDoc to generate documentation. 

Ideally, you would want to disable this rule near the the offending line in the comment, like this:

```js
test.cb('foo', t => {
  // My comments might describe test code, like:
  //
  // We have to use the `test.cb()` form for this test because... // eslint-disable-line
  //
  doSomething(t.end);
});
```

Not sure how disable comments are attached / handled, but I would imaging they attach to actual Nodes (not comments). How does it apply to leading / trailing comments? Need to play with it some more.

rules/no-commented-test.js
```diff
@@ -0,0 +1,112 @@
...
+					locationInComments.column
+			};
+		}
+
+		linesLeft -= linesThisComment;
+	}
+
+	// TODO: Probably should not throw here. Maybe return `null` and modify the message warning the exact location could not be found?
+	// This should be unreachable code, so would like to encourage bug reports if that ever proves untrue.
+	throw new Error('unable to translate location');
```
sindresorhus: I think throwing is fine if this is meant to be never hit. The error message should improved though. It should say that this should really never happen and tell the user to open a bug report with the link `https://github.com/sindresorhus/eslint-plugin-ava/issues/new`.

rules/no-commented-test.js
```diff
@@ -0,0 +1,112 @@
...
+}
+
+function isTestFunctionCall(node) {
```
sindresorhus: Move this to `util.js`?

test/no-commented-tests.js
```diff
@@ -0,0 +1,48 @@
...
+			{
+				code: header + 'test.cb(t => { t.pass(); });\n// test.cb(t=> {})',
+				errors: [errorAt(3, 4)]
+			},
+			{
+				code: header + 'test.cb(t => { t.pass(); });\n // test.cb(t=> {})',
+				errors: [errorAt(3, 5)]
+			},
+			{
+				code: header + 'test.cb(t => { t.pass(); });\n//  test.cb(t=> {})',
```
sindresorhus: Space after `t` on all of these.

test/no-commented-tests.js
```diff
@@ -0,0 +1,48 @@
...
+				errors: [errorAt(3, 5)]
+			},
+			{
+				code: header + 'test.cb(t => { t.pass(); });\n// This is not javascript!\n// test.cb(t=> {})',
+				errors: [errorAt(4, 4)]
+			}
```
sindresorhus: I would like to see some more tests.

- Multi-line comment
- `test(t =>`

sindresorhus:
Looks promising.

As for naming, we should probably follow our convention (e.g. `no-only-test`) of singular `test` and name it `no-commented-test`?


> It will likely be faster to first scan the comment block with a regular expression

:+1:

> Should we maybe require valid JS between the opening

:+1:

> Especially if you are using JSDoc to generate documentation.

Can't we just ignore JSDoc comments? Comments starting with `/**`.

jamestalmage:
Ah yes - ignoring jsdoc. That should work.

revelt:
hi guys! Sorry to pester but I'm really keen to use this feature. Where are we stuck?

sindresorhus:
@revelt It's all in the feedback comments.

sindresorhus:
Closing for lack of activity.

PR #86, opened 2016-04-17
[WIP]: add utils for checking `test` arg types

jamestalmage:
This started as something I was considering for #85 (no commented out tests).

I figured it would be good to allow users to comment out lines where they were passing arguments we knew were illegal (i.e. only a string to `test('str')` - though maybe we should tell them to uncomment it and use `todo`). 

I'm thinking maybe this has utility across some of the other rules, and could serve as the basis for an additional rule that flagged incorrect arguments to `test(...)`

Right now it's just two utility methods: `isValidTestName` and `isValidTestFunction`. Really, those could just be renamed `isString` and `isFunction` respectively, though we may go further with `isValidTestFunction` by evaluating it's parameters. 

It doesn't do much static type analysis (it never examines the scope), it just checks if it's a StringLiteral, etc.

The methods return one of `"definitely"`, `"maybe"`, or `false`. A `StringLiteral` argument is `definitely` a valid test name argument. An identifier can be anything, so it's a `maybe`. `false` is returned for constructs where we can determine the exact type, and it's the wrong one.

### DIFF
--- a/test/utils.js
+++ b/test/utils.js
@@ -1,0 +1,87 @@
+import test from 'ava';
+import babelEslint from 'babel-eslint';
+import utils from '../rules/utils';
+
+// isValidTestName
+
+function testName(code, expected) {
+	test(`isTestName(${code}) === $expected`, t => {
+		var node = babelEslint.parse(code).body[0].expression;
+		t.is(utils.isValidTestName(node), expected);
+	});
+}
+
+testName('"string"', 'definitely');
+testName('`template`', 'definitely');
+
+testName('callExp()', 'maybe');
+testName('this', 'maybe');
+testName('identifier', 'maybe');
+testName('obj.prop', 'maybe');
+
+test('YieldExpression is "maybe" a valid test name', t => {
+	var node = babelEslint.parse('function* foo() {yield b}').body[0].body.body[0].expression;
+	t.is(utils.isValidTestName(node), 'maybe');
+});
+
+testName('a += "str"', 'definitely');
+testName('a = "str"', 'definitely');
+testName('a += b', 'maybe');
+testName('a = b', 'maybe');
+
+testName('id1 + id2', 'maybe');
+testName('id1 + call()', 'maybe');
+testName('call() + id2', 'maybe');
+testName('call1() + call2()', 'maybe');
+testName('"str" + call2()', 'definitely');
+testName('id1 + "str"', 'definitely');
+testName('"foo" - "foo"', false);
+testName('3 + 4', false);
+
+testName('condition ? id1 : id2', 'maybe');
+testName('condition ? id1 : "string2"', 'maybe');
+testName('condition ? "string1" : "string2"', 'definitely');
+testName('condition ? 1 : "string2"', false);
+
+testName('a || b', 'maybe');
+testName('"a" || "b"', 'definitely');
+testName('a || "b"', 'maybe');
+testName('a && b', false);
+
+// isValidTestFunction
+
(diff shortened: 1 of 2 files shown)


test/utils.js
```diff
@@ -0,0 +1,87 @@
+import test from 'ava';
+import babelEslint from 'babel-eslint';
+import utils from '../rules/utils';
+
+// isValidTestName
+
+function testName(code, expected) {
+	test(`isTestName(${code}) === $expected`, t => {
+		var node = babelEslint.parse(code).body[0].expression;
```
sindresorhus: `const`

test/utils.js
```diff
@@ -0,0 +1,87 @@
...
+testName('"string"', 'definitely');
+testName('`template`', 'definitely');
+
+testName('callExp()', 'maybe');
+testName('this', 'maybe');
+testName('identifier', 'maybe');
+testName('obj.prop', 'maybe');
+
+test('YieldExpression is "maybe" a valid test name', t => {
+	var node = babelEslint.parse('function* foo() {yield b}').body[0].body.body[0].expression;
```
sindresorhus: `const`

test/utils.js
```diff
@@ -0,0 +1,87 @@
...
+testName('a || b', 'maybe');
+testName('"a" || "b"', 'definitely');
+testName('a || "b"', 'maybe');
+testName('a && b', false);
+
+// isValidTestFunction
+
+function testFn(code, expected) {
+	test(`isTestFunction(${code}) === $expected`, t => {
+		var node = babelEslint.parse('(' + code + ')').body[0].expression;
```
sindresorhus: `const`

sindresorhus:
Good idea!

We already have a `util.js` file, so maybe create a `util` folder?

jfmengels:
Maybe we can create a new repo for it, that way it can be improved over time and several plugins can make use of it.

In this repo, we have `create-ava-rule` that has a `merge` function and allows to apply a Visitor function before/after the visitor implemented in the rule, and keep some things in memory (is this a test file, is current function an implementation function, ...?).

I use a similar approach in `eslint-plugin-lodash-fp` to remember whether Lodash is imported, how, what identifier correspond to a Lodash function, etc., again with a `merge` functionality.

I started creating a separate module (`eslint-enhanced-visitors`, haven't pushed it to GH yet actually) that created a merge function, and that can serve as memory for those kind of applications.

So, I was thinking we could have another set of visitors (in a separate module) that could be merged, and that remembers the types of what it could deduce, so that it becomes available to other plugins.

I thought it would be something like this:
```js
// type-computer module
{
  VariableDeclaration: function (node) { ... },
  ...
};

// create-ava-rule.js
var enhance = require('eslint-enhanced-visitors');
var typeComputer = require('eslint-type-computer');
module.exports = function() {
  merge: function(visitorSet) {
    return enhance([typeComputer, avaVisitor, visitorSet]);
  }
};

// some-rule.js
module.exports = function() {
  var info = createAvaRule();

  return info.merge({
    CallExpression: function(node) {
      if (info.type[callee.name] === 'string') {
        // ...
      }
    }
  });
};
```

sindresorhus:
I really like the idea of reuse. Right now pretty much every ESLint plugin reinvents utils.

sindresorhus:
Let's merge this here for now and someone can extract it later. @jamestalmage What's left before being mergeable? Anything you'd like feedback on?

sindresorhus:
Closing for lack of activity.

PR #89, opened 2016-04-26
Update Node.js version for Travis

jfmengels:
Update Node.js version for Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - '5'
+  - '6'
   - '4'
   - '0.12'
   - '0.10'


jfmengels:
Not that surprising, but it works on v6, yay :)

ntwb:
Should this change `5` to `6` rather than just adding `6`?

It is really just a major semver version bump than a new release branch IMHO.

See also: NodeJS as part of 6.0.0 have renamed `stable` to `current` in [nodejs/node/pull/6318](https://github.com/nodejs/node/pull/6318)

sindresorhus:
Yes, let's drop testing on Node.js 5. We're only interested in testing on current and LTS releases.

> Node.js v5 will continue to be supported for the next two months in order to give developers currently using v5 time to transition to Node.js v6. - https://nodejs.org/en/blog/release/v6.0.0/

novemberborn:
@sindresorhus while I doubt there'd be regressions that affect *just* v5, it seems sensible to keep testing until official support is withdrawn. Especially in the first few weeks when people won't necessarily have upgraded.

> It is really just a major semver version bump than a new release branch IMHO.

@ntwb it's a major bump because of the breaking changes, which weren't previously in v5.

PR #90, opened 2016-04-29
Properly check for valid argument Node in use-true-false rule

bojand:
This fixes a crash in Atom using `linter-xo` that happens while typing on the opening bracket of `t.truthy(` or when closing on empty argument. This check makes sure the `arg` value is completely taken into account within the check before proceeding to check any of the `||` combinations. I was getting the crash in `eslint-plugin-ava` shipped with `xo@0.14` which has different code than the master here; but I think the fix is still valid here.

### DIFF
--- a/rules/use-true-false.js
+++ b/rules/use-true-false.js
@@ -64,10 +64,10 @@ module.exports = function (context) {
 				var arg = node.arguments[0];
 
 				if (arg &&
-					(arg.type === 'BinaryExpression' && booleanBinaryOperators.indexOf(arg.operator) !== -1) ||
+					((arg.type === 'BinaryExpression' && booleanBinaryOperators.indexOf(arg.operator) !== -1) ||
 					(arg.type === 'UnaryExpression' && arg.operator === '!') ||
 					(arg.type === 'Literal' && arg.value === Boolean(arg.value)) ||
-					(matchesKnownBooleanExpression(arg))
+					(matchesKnownBooleanExpression(arg)))
 				) {
 					if (node.callee.property.name === 'falsy') {
 						context.report({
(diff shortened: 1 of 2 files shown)


jfmengels:
Ah yes, that makes total sense. 

Would you mind adding new test cases for this? `testCase('t.truthy()')` and `testCase('t.falsy()')`. You can put them in the `valid` section I'd say (speaking of which, maybe we should a have a rule saying that assertions should always have arguments?)

Thanks a lot @bojand!

bojand:
Added tests. Should have checked and done it from start.

Yea that may be a good idea to have something like `no-empty-arg` rule. You do mean like a separate rule altogether correct, and nothing related to this PR? Or specifically within context of this rule? Are there valid scenarios perhaps where someone would want such usage? I can't really think of it but I don't know. If it's a separate rule I imagine that can be a different PR altogether, its kind of outside of the scope of this.

jfmengels:
> You do mean like a separate rule altogether correct, and nothing related to this PR?

Yes ;)

> Are there valid scenarios perhaps where someone would want such usage?

Having an assertion without arguments? I don't think so no. Obviously there should be an exception for `t.end()`, `t.pass()` and `t.fail()`, but for the others, I don't see how it would be a valid use.

LGTM :)

sindresorhus:
> Yea that may be a good idea to have something like no-empty-arg rule.

Instead, how about a rule that ensures the argument length is correct for each assertion?

So `t.pass()` would be allowed, `t.pass('msg')` too, but not `t.pass('msg, 'foo')`.

sindresorhus:
Thank you @bojand :)

jfmengels:
https://github.com/eslint/eslint/issues/6023 :)

PR #94, opened 2016-05-01
Add tests for checking the index file

jfmengels:
Add tests for checking the index file

I've been bitten a few times in another my lodash-fp plugin by the fact that my `index.js` file had some typos in it (either in the exported rule, or the recommended config).

I've thus added a test to verify those.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "eslint": "^2.2.0",
     "js-combinatorics": "^0.5.0",
     "nyc": "^6.4.0",
+    "pify": "^2.3.0",
     "xo": "*"
   },
   "peerDependencies": {
--- a/test/package.js
+++ b/test/package.js
@@ -1,0 +1,20 @@
+import fs from 'fs';
+import test from 'ava';
+import pify from 'pify';
+import index from '../';
+
+test('Every rule is defined in index file', async t => {
+	const files = await pify(fs.readdir, Promise)('../rules/');
+	const rules = files.filter(file => file.indexOf('.js') === file.length - 3);
+
+	rules.forEach(file => {
+		const name = file.slice(0, -3);
+		t.truthy(index.rules[name], `'${name}' is not exported in 'index.js'`);
+		t.truthy(index.configs.recommended.rules[`ava/${name}`], `'${name}' is not set in the recommended config`);
+	});
+
+	t.is(Object.keys(index.rules).length, rules.length,
+		'There are more exported rules than rule files.');
+	t.is(Object.keys(index.configs.recommended.rules).length, rules.length,
+		'There are more exported rules in the recommended config than rule files.');
+});


test/package.js
```diff
@@ -0,0 +1,24 @@
+import fs from 'fs';
+import test from 'ava';
+import index from '../';
+
+test.cb('Every rule is defined in index file', t => {
+	fs.readdir('../rules/', function (err, files) {
```
sindresorhus: Can you use https://github.com/sindresorhus/pify to promisify this?

test/package.js
```diff
@@ -0,0 +1,24 @@
...
+		t.ifError(err);
+
+		var rules = files.filter(file => file.indexOf('.js') === file.length - 3);
```
sindresorhus: `const`

test/package.js
```diff
@@ -0,0 +1,24 @@
...
+
+		rules.forEach(file => {
+			const name = file.slice(0, -3);
+			t.truthy(index.rules[name], `'${name}' is not exported in 'index.js'`);
+			t.truthy(index.configs.recommended.rules['ava/' + name], `'${name}' is not set in the recommended config`);
```
sindresorhus:
```
`ava/${name}`
```

sindresorhus:
Good idea!

test/package.js
jfmengels: Wondering about something... why is `esnext` not activated in XO for test files? You can activate rules for a glob right? Why not activate it for things in the default AVA test glob?

sindresorhus: I don't like project specific config, and we're very close to being able to just target Node.js 4 (next ESLint version), so didn't want to overhead.

sindresorhus: Feel free to do it if you care enough, but it will be moot in a couple of months.

jfmengels: Nah it's fine, just wanted to have your opinion on it.

test/package.js
```diff
@@ -0,0 +1,22 @@
+import fs from 'fs';
+import test from 'ava';
+import pify from 'pify';
+import Promise from 'pinkie-promise';
```
sindresorhus: You don't need this. `global.Promise = Promise;` works fine since Babel replaces the right part with a polyfill.

package.json
```diff
@@ -75,6 +75,8 @@
     "eslint": "^2.2.0",
     "js-combinatorics": "^0.5.0",
     "nyc": "^6.4.0",
+    "pify": "^2.3.0",
+    "pinkie-promise": "^2.0.0",
```
sindresorhus: Not needed.

test/package.js
```diff
@@ -0,0 +1,22 @@
...
+import index from '../';
+global.Promise = Promise;
```
sindresorhus: Btw, you can pass `Promise` as the second argument in `pify`.

jfmengels: Ah, yes, will do that :)

sindresorhus:
👍

jfmengels:
I'll probably add the same test file to `eslint-plugin-xo` (I have added it to `eslint-plugin-lodash-fp` too). Wondering if this is an "okay" copy/paste code, or if this should be a module (doesn't seem generic enough to me)

sindresorhus:
Yeah, sure, doesn't seem generic enough for a module.

PR #95, opened 2016-05-01
Add `assertion-arguments` rule (fixes #92)

### LINKED ISSUE #92: Rule to enforce the correct number of arguments for assertions
As discussed in https://github.com/sindresorhus/eslint-plugin-ava/pull/90#issuecomment-215694147 and in https://github.com/sindresorhus/eslint-plugin-ava/pull/90#issuecomment-215832722.

jfmengels:
Add `assertion-arguments` rule (fixes #92)

Willing to change the name of the rule, but this is pretty generic, and maybe some day we'll want to detect the argument types. 

Probably to be followed by the deprecation of `assertion-message`.

### DIFF
--- a/docs/rules/assertion-arguments.md
+++ b/docs/rules/assertion-arguments.md
@@ -1,0 +1,62 @@
+# Enforce passing correct arguments to assertions
+
+Enforces passing the right number of arguments to assertion methods like `t.is()`. This rule can optionally also enforce or forbid the use of assertion messages.
+
+Assertion messages are optional arguments that can be given to any assertion call to improve the error message, should the assertion fail.
+qq
+
+## Fail
+
+```js
+import test from 'ava';
+
+test(t => {
+	t.is(value); // Not enough arguments
+	t.is(value, expected, message, extra); // Too many arguments
+});
+
+/* eslint ava/assertion-arguments: ["error", {"message": "always"}] */
+test(t => {
+	t.true(array.indexOf(value) !== -1);
+});
+
+/* eslint ava/assertion-arguments: ["error", {"message": "never"}] */
+test(t => {
+	t.true(array.indexOf(value) !== -1, 'value is not in array');
+});
+```
+
+
+## Pass
+
+```js
+import test from 'ava';
+
+test(t => {
+	t.is(value, expected);
+	t.is(value, expected, message);
+});
+
+/* eslint ava/assertion-arguments: ["error", {"message": "always"}] */
+test(t => {
+	t.true(array.indexOf(value) !== -1, 'value is not in array');
+});
+
+/* eslint ava/assertion-arguments: ["error", {"message": "never"}] */
+test(t => {
+	t.true(array.indexOf(value) !== -1);
+});
+```
+
+
+## Options
+
+This rule supports the following options:
+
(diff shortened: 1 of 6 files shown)


docs/rules/assertion-arguments.md
```diff
@@ -0,0 +1,59 @@
...
+
+/* eslint ava/assertion-arguments: ["error", "never"] */
+test(t => {
+	t.true(array.indexOf(value) !== -1);
+});
+```
+
+## Options
+
+The rule takes one option, a string, which could be either `"always"` (enforces error messages) or `"never"` (forbids assertion messages). If omitted or set to false, assertion messages will be neither enforced nor forbidden.
```
sindresorhus: It no longer makes sense to have this is a main boolean option. I think it should rather be part of an options object.

sindresorhus:
Maybe we should also ensure the message argument is a string? I know you could potentially put it in an external string, but most put it inline, so we could just de-opt on non-literal values. Or maybe even use code paths as if it's not inline, it's always in the same file.

jfmengels:
Yeah, I was thinking the same thing, and also checking that `n` in `t.plan(n)` is a integer. Wanted to wait before I started working on a shared analysis tool, but this can be simple, so as you wish.

If we need to play with code paths, I'd rather put it as a follow-up tweak, still can't figure my head around it ^^'

jfmengels:
Updated

docs/rules/assertion-arguments.md
```diff
@@ -0,0 +1,61 @@
...
+	t.is(value); // Not enough arguments
+	t.is(value, expected, message, extra); // Too many arguments
+});
+
+/* eslint ava/assertion-arguments: ["error", "always"] */
+test(t => {
+	t.true(array.indexOf(value) !== -1);
+});
+
+/* eslint ava/assertion-arguments: ["error", "never"] */
```
sindresorhus: These comments needs to be updated with the new format.

docs/rules/assertion-arguments.md
```diff
@@ -0,0 +1,61 @@
...
+test(t => {
+	t.true(array.indexOf(value) !== -1, 'value is not in array');
+});
+
+/* eslint ava/assertion-arguments: ["error", "never"] */
+test(t => {
+	t.true(array.indexOf(value) !== -1);
+});
+```
+
```
sindresorhus: One more newline

docs/rules/assertion-arguments.md
jfmengels: Ugh... yes.

docs/rules/assertion-arguments.md
jfmengels: The more I write ESLint rules doc, the more I want to build a markdown and JSON linter...

jfmengels:
Tada!

docs/rules/assertion-arguments.md
sindresorhus: You should talk to @wooorm. He has already done that. https://github.com/wooorm/remark-lint

jfmengels:
Updated

docs/rules/assertion-arguments.md
jfmengels: Will check it out!

sindresorhus:
Splendid work as always @jfmengels 🙌

leo-cat

jfmengels:
Hey, a DiCaprio gif, it's been a while :D

sindresorhus:
I recently stocked up on more of those.

PR #97, opened 2016-05-04
fix use-t rule crashing when test() has no arguments

nfcampos:
`use-t` rule crashes with the snippet below, which obviously is not a proper test file but it's what the linter sees if you run it as you type (with eg. the `atom` plugin)

```js
import test from 'ava'

test()
```

I've fixed it and added a test for it.

### DIFF
--- a/rules/use-t.js
+++ b/rules/use-t.js
@@ -12,7 +12,7 @@ module.exports = function (context) {
 
 			var functionArg = node.arguments[node.arguments.length - 1];
 
-			if (!functionArg.params || !functionArg.params.length) {
+			if (!(functionArg && functionArg.params && functionArg.params.length)) {
 				return;
 			}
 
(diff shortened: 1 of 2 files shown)


rules/use-t.js
```diff
@@ -12,7 +12,7 @@ module.exports = function (context) {
 
 			var functionArg = node.arguments[node.arguments.length - 1];
 
-			if (!functionArg.params || !functionArg.params.length) {
+			if (!functionArg || !functionArg.params || !functionArg.params.length) {
```
sindresorhus: With so many negated arguments, I think this would be clearer:

```js
if (!(functionArg && functionArg.params && functionArg.params.length)) {
```

nfcampos: Good point, I've changed it

sindresorhus:
Thank you :)

jfmengels:
Thanks a lot @nfcampos!

sindresorhus:
New version published.

PR #100, opened 2016-05-04
`assertion-arguments`: Allow third argument for `t.throws`. Fixes #99

### LINKED ISSUE #99: Assertion Arguments warns t.throws with 3 arguments
**eslint version:** 2.9.0
**eslint plugin ava:** 2.3.0

`assertion-arguments` rule warns when `t.throws` is passed three parameters.
According to [AVA docs](https://github.com/sindresorhus/ava#throwsfunctionpromise-error-message) `t.throws` is valid with one, two, or three parameters.

.eslintrc.yml
``` yml
env:
    node: true
    es6: true
plugins:
    - ava
rules:
    ava/assertion-arguments: error
```

example code
```js
const test = require('ava');

test('reduced test case', (t) => {
    // valid test
    // passes lint check
    t.throws(
        () => {
            throw new Error('example');
        }
    );

    // valid test
    // passes lint check
    t.throws(
        () => {
            throw new Error('example');
        },
        'example'
    );

    // valid test
    // fails lint check
    t.throws(
        () => {
            throw new Error('example');
        },
        'example',
        'some message'
    );
});
```

jfmengels:
`assertion-arguments`: Allow third argument for `t.throws`. Fixes #99


sindresorhus:
Wow. That was quick :)

jfmengels:
Yep. Not the right decision sleep-wise though ^^'

PR #102, opened 2016-05-06
Support `always` modifier in `no-unknown-modifiers`

jfmengels:
With regards to the work being done in https://github.com/sindresorhus/ava/pull/806.
Since it looked like it will be merged soon, I've prepared the work here.


sindresorhus:
Looks good. Let's wait until it lands in AVA.

PR #103, opened 2016-05-09
Fix typo assertion-arguments

forresst:


jfmengels:
Oops. Thanks @forresst!

Issue #104, opened 2016-05-09
Improve test failures with RuleTester and AVA


jfmengels:
When developping and running the tests, I often get the following:

![image](https://cloud.githubusercontent.com/assets/3869412/15127142/8edf39de-1635-11e6-8e19-b3c8981fee04.png)

As you can see and imagine, this error message is really not helpful, and I often have to uncomment all the tests to find out which one fails. I thought this was the normal behavior of `RuleTester`, but then I started contributing to https://github.com/benmosher/eslint-plugin-import, where the errors are a bit clearer
![image](https://cloud.githubusercontent.com/assets/3869412/15127326/72f9d3fe-1636-11e6-8d11-f425f07deeac.png)

I have the same problem with an other eslint plugin repo of mine, so it would be pretty neat if this could be improved. Do you guys know of ways to improve it?

twada:
Stop writing everything in a single `test` function. It's an [Assertion Roulette](http://xunitpatterns.com/Assertion%20Roulette.html) antipattern.

To improve, see [test/prefer-power-assert.js](https://github.com/sindresorhus/eslint-plugin-ava/blob/9eaf1cde7f4789cc2f59a7627a1c21c0d827a2bd/test/prefer-power-assert.js)

sindresorhus:
@jfmengels See: https://github.com/sindresorhus/ava/issues/764#issuecomment-210701728

jfmengels:
Man, RuleTester was really built with Mocha in mind...

![image](https://cloud.githubusercontent.com/assets/3869412/15192832/36b27622-17bb-11e6-8a09-ac46bfb61b2f.png)

versus

![image](https://cloud.githubusercontent.com/assets/3869412/15192864/60fab624-17bb-11e6-9758-53963e816eb9.png)

Even if you wrap every single one of them in a seperate test, and add https://github.com/sindresorhus/ava/issues/764#issuecomment-210701728 (though that will help a lot with the kind of error I've shown here), the built-in feedback is much clearer (but you lose the transpiling)... Meh :(

sindresorhus:
How is the Mocha output clearer when https://github.com/sindresorhus/ava/issues/764#issuecomment-210701728 is fixed? What can we improve?

Also keep in mind that you can use mocha reporters with the `--tap` option and [`tap-mocha-reporter`](https://github.com/tapjs/tap-mocha-reporter).

jfmengels:
This it the output with the mocha reporter (for those interested):
[...]

PR #105, opened 2016-05-11
Improve tests feedback, linked to #104

jfmengels:
This is related to #104, and an attempt to improve the RuleTester's feedback. By default, it uses Mocha's `describe` and `it` to create rules, but it they are not defined, it uses the functions [here](https://github.com/eslint/eslint/blob/master/lib%2Ftesters%2Frule-tester.js#L436-L452).

I've managed to override the RuleTester test creation a bit to use AVA's `test` rather than Mocha's `it`, and here are the results:

### When run with Mocha
image

### When run with AVA (with `--verbose`)
image

### When run with AVA and `tap-mocha-reporter`

image

The odd `test.xxx(t => {})` kind of everywhere are the second part of the title (since it takes as title the code, and the code has a `\n` in the middle). If we remove the `\n`, it gets much clearer
image

Please check out [`eslint-ava-rule-tester`](https://github.com/jfmengels/eslint-ava-rule-tester), which exports a modified RuleTester. I'm up for modifying the name, but it's already published (could not make it all work with `npm link`), which contains the same code that you can see in `test/helpers/ava-rule-tester.js` (I kept it ll so that you guys could easily see what I did here).

Let me know what you guys think.

PS: Obviously this is just a draft, I haven't migrated all the tests and have even left an error so that you can see the error feedback more easily.


sindresorhus:
Nice! Looks so much better now :)

sindresorhus:
Maybe we should do a PR to ESLint adding a hook to their rule checker so any test framework could easily hook into it?

jfmengels:
> Nice! Looks so much better now :)

Yeah!

> Maybe we should do a PR to ESLint adding a hook to their rule checker so any test framework could easily hook into it?

Sure. By their contribution guidelines, we should first create an issue though. Not sure I see what the hook should look like though?

jfmengels:
By the way: Found this comment https://github.com/eslint/eslint/issues/5040#issuecomment-182693246, where they talk about eventually using AVA under the hood for RuleTester.

jfmengels:
Created an issue in ESLint about this https://github.com/eslint/eslint/issues/6227

jfmengels:
Now that we know that ESLint considers this okay and `describe` and `it` as part of their API (meaning it will be enforced by semver), shall I go ahead with finishing this? If so, I'll go ahead and do `eslint-plugin-xo` too (and my own plugins).

jfmengels:
### Before
image

### After
image

Biggest number of test additions you'll ever see in one commit :D

jfmengels:
Hmm, looking into the tests a bit later today. Stay tuned.

sindresorhus:
> shall I go ahead with finishing this?

Let's do it! :D

jfmengels:
There you go :)

sindresorhus:
Yay! So much better output :)

leo-excited

jfmengels:
:)

Will do a PR for xo soon-ish.

jfmengels:
PS: https://github.com/eslint/eslint/pull/6331, if you have comments to improve this ;)

PR #106, opened 2016-05-15
Allow multiple anonymous rules in `no-identical-title`

kevva:
Being `undefined`, I guess we can assume it's a anonymous title.

### DIFF
--- a/rules/no-identical-title.js
+++ b/rules/no-identical-title.js
@@ -12,8 +12,7 @@ function isStaticTemplateLiteral(node) {
 }
 
 function isStatic(node) {
-	return !node ||
-		node.type === 'Literal' ||
+	return node.type === 'Literal' ||
 		(node.type === 'TemplateLiteral' && isStaticTemplateLiteral(node)) ||
 		(node.type === 'BinaryExpression' && isStatic(node.left) && isStatic(node.right));
 }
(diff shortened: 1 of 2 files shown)


rules/no-identical-title.js
```diff
@@ -39,7 +38,7 @@ module.exports = function (context) {
 			var args = node.arguments;
 			var titleNode = args.length > 1 || ava.hasTestModifier('todo') ? args[0] : undefined;
 
-			if (!isStatic(titleNode)) {
+			if (titleNode === undefined || !isStatic(titleNode)) {
```
kevva: Checking explicitly here so we don't treat empty strings or whatever as anonymous tests.

test/no-identical-title.js
```diff
@@ -29,6 +29,8 @@ test(() => {
 			header + 'test.beforeEach(t => {}); test.beforeEach(t => {});',
 			header + 'test.afterEach(t => {}); test.afterEach(t => {});',
 			header + 'test.cb.before(t => {}); test.before.cb(t => {});',
+			header + 'test(t => {}); test(t => {});',
```
jamestalmage: Maybe add a comment here - "multiple anonymous tests covered by the if-multiple rule"

rules/no-identical-title.js
jamestalmage: Another comment here:

> Don't flag computed titles or anonymous tests (anon tests covered in the if-multiple rule)

test/no-identical-title.js
kevva: Good idea.

jamestalmage:
LGTM

sindresorhus:
LGTM2

Thanks @kevva :)

PR #111, opened 2016-05-18
Simplifying tests

### LINKED ISSUE #72: Make writing "within a test" and "t.XXX" matchers easier:
We have this code construct everywhere:

```js
CallExpression: function (node) {
  // only check if we are within a test file, and if this CallExpression is within a test.
  if (!ava.isTestFile || !ava.currentTestNode) {
    return;
  }
}
```

I propose we add some sugar to `ava.merge()` that allows something like this:

```js
'CallExpression:inTest': function (node) {

},
'VariableDeclaration:inTestFile': function (node) {

},
'CallExpression:inTest:inTestFile': function (node) {

},

```

While we are at it, we should define one (specifically for CallExpression), that checks the callee is a `MemberExpression` of the form `t.XXX`:

```js
'AssertionCallExpression:inTest': function (node) {

}
```

jfmengels:
Fixes #72
WIP, just tackled one of the rules, to see the reactions. Does this still look good to you @sindresorhus @jamestalmage @twada? If so, I'll go and do the others.

### DIFF
--- a/create-ava-rule.js
+++ b/create-ava-rule.js
@@ -1,6 +1,7 @@
 'use strict';
 var espurify = require('espurify');
 var assign = require('object-assign');
+var rest = require('lodash.rest');
 var deepStrictEqual = require('deep-strict-equal');
 
 var avaImportDeclarationAst = {
@@ -102,11 +103,11 @@ module.exports = function createAvaRule() {
 		hasTestModifier: function (mod) {
 			return hasTestModifier(currentTestNode, mod);
 		},
-		hasHookModifier: function () {
-			return hasTestModifier(currentTestNode, 'before') ||
-				hasTestModifier(currentTestNode, 'beforeEach') ||
-				hasTestModifier(currentTestNode, 'after') ||
-				hasTestModifier(currentTestNode, 'afterEach');
+		hasNoHookModifier: function () {
+			return !hasTestModifier(currentTestNode, 'before') &&
+				!hasTestModifier(currentTestNode, 'beforeEach') &&
+				!hasTestModifier(currentTestNode, 'after') &&
+				!hasTestModifier(currentTestNode, 'afterEach');
 		},
 		merge: function (customHandlers) {
 			Object.keys(predefinedRules).forEach(function (key) {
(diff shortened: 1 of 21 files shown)


jamestalmage:
Yes. This is fantastic.

sindresorhus:
👌

jfmengels:
There you go. I think we can improve it a bit further by handling cases like `isCalleeMemberExpression` or `hasTestModifier`, but this is at least a good start and handles the most common case.

create-ava-rule.js
```diff
@@ -129,16 +130,29 @@ module.exports = function createAvaRule() {
...
-		}
+	rule.isTestNode = function (node) {
+		return currentTestNode === node;
+	};
+
+	rule.isInTestNode = function () {
+		return currentTestNode;
+	};
+
+	rule.if = rest(function (predicates) {
```
twada: [nits] `if` is a reserved word for ECMAScript so we better avoid it.

Good alternatives? `when` ?

jamestalmage: That's only needed for ES3 backwards compatibility.

https://mathiasbynens.be/notes/javascript-properties

twada: Yep, so it's just a nitpicking.
But it also make syntax highlighting annoying IMO.

jfmengels: Yeah, colorwise, it gets shown with the same color as an if in m'y editor, and I rhoight that was actually pretty convenient, as that translates pretty well what it does. Happy to change it to when if you want..

sindresorhus: I prefer `if`. We don't care about ES3. If your editor is highlighting it incorrectly, I'd suggest opening and issue/PR on your editor.

Sublime is very open to pull requests for syntax highlighting now btw: https://github.com/sublimehq/Packages

jfmengels: In Atom for instance, the `return` method of Bluebird
```js
return Promise.resolve(foo)
  .return(bar);
```

also gets highlighted like the `return` keyword. Which is nice IMO :)

twada: Agreed to go with `if` 👍

twada:
LGTM 👍

jfmengels:
season-4-the-simpsons-4x2-l2JdVRfJozpjq70SA

PR #112, opened 2016-05-22
Allow macro use with `test-title` rule.

jamestalmage:
Currently, the `test-title` rule will freak out if you use macros with more than one argument.

Unfotuneately, this simple fix is going to let some errors slip through. Specifically, if you don't attach a `title` function to your macro, and don't supply a string title, we probably should error, but that requires some rather involved type analysis to see if they ever did attach a title method to the macro.

### DIFF
--- a/rules/test-title.js
+++ b/rules/test-title.js
@@ -15,7 +15,7 @@ module.exports = function (context) {
 			testCount++;
 
 			var requiredLength = ava.hasTestModifier('todo') ? 1 : 2;
-			var hasNoTitle = node.arguments.length !== requiredLength;
+			var hasNoTitle = node.arguments.length < requiredLength;
 			var isOverThreshold = !ifMultiple || testCount > 1;
 
 			if (hasNoTitle && isOverThreshold) {
--- a/test/test-title.js
+++ b/test/test-title.js
@@ -64,6 +64,10 @@ test(() => {
 				code: header + 'notTest(t => { t.pass(); t.end(); });',
 				options: ['always']
 			},
+			{
+				code: header + 'test(macroFn, arg1, arg2);',
+				options: ['always']
+			},
 			// shouldn't be triggered since it's not a test file
 			{
 				code: 'test(t => {});',


sindresorhus:
LGTM

jfmengels:
LGTM. And yeah, I really like the idea of macros, but this is going to make linting much harder for several rules.

jamestalmage:
Should we have done `test.macro(macroFn, arg, arg)`?

I know you proposed that at one point.

I guess I don't see it being a whole lot more difficult. If they are not declaring the test implementation inline, it's almost certainly a macro, and even if it's not - that was always going to be difficult to analyze.

jfmengels:
Nah, `test(macroFN)` is really good IMO. But we need to up our game to detect what function is an implementation function when not used inline with `test`, and using `test.macro` would not help.

For instance, this will be hard to detect (is `t.end()` valid, should it necessary have `t.end()`, ...?)
```js
function impl(t, arg) {
  t.end();
}

test.cb(impl, 1);
test(impl, 1);
```

sindresorhus:
@jfmengels Why is it hard to detect?

jfmengels:
Well, not necessarily hard, but harder, and currently not supported.

For `no-invalid-end` (no `t.end()` outside of `test.cb`), we check whether we're currently in a non-`test.cb()` call, and whether at the end of one, a `t.end` has been seen.
If you want to support macros for it, you'd have to analyse every function and see if it mentions `t.end`, store them all in some object (by function name? Hoping it has one, and it never gets overriden), store all the `test()` calls with that function and then in a `Program:exit` visitor, check that every non-cb test call uses an implementation for which we have found that there was no `t.end`.

The `test-ended` rule currently reports an error for `test.cb(fn)`. If we want to support that as well, we'd have to go find that function and then analyse it. Meaning, in practice, we have to store all the 

I'm pretty sure that none of our assertion rules will check anything outside of an implementation function outside of a `test()` call, so the following is not reported for instance.

```js
function impl(t) {
  t.foobaridontexist();
}

test(impl);
```

Maybe we should simply report for functions that have t as their first argument.

jamestalmage:
> Maybe we should simply report for functions that have t as their first argument.

I think that is worth exploring. We could use `ava-files` to only apply the rule in tests and helper files.

PR #118, opened 2016-06-01
Update prefer-power-assert to accept falsy, true, false

### LINKED ISSUE #117: `prefer-power-assert` does not allow the use of `t.true()`
Since `t.true` works with power-assert, the rule should not warn when using it. I think it should also allow `t.false` and `t.falsy` as well.

In the gitter chat, @sindresorhus [suggested](https://gitter.im/avajs/ava?at=574e861f6bbc2d1d4df00e6a) that `prefer-power-assert` should enforce `t.true` over `t.truthy`, but that seems to me like that is covered by the `use-true-false` rule.

Happy to submit a PR with the changes.

bdougherty:
Fixes #117

### DIFF
--- a/docs/rules/prefer-power-assert.md
+++ b/docs/rules/prefer-power-assert.md
@@ -2,13 +2,13 @@
 
 Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/related/eslint-plugin-ava/docs/rules/prefer-power-assert.md)
 
-- [`t.truthy()`](https://github.com/sindresorhus/ava#truthyvalue-message) __(You can do [most things](https://github.com/sindresorhus/ava#enhanced-assertion-messages) with this one)__
-- [`t.deepEqual()`](https://github.com/sindresorhus/ava#deepequalvalue-expected-message)
-- [`t.notDeepEqual()`](https://github.com/sindresorhus/ava#notdeepequalvalue-expected-message)
-- [`t.throws()`](https://github.com/sindresorhus/ava#throwsfunctionpromise-error-message)
-- [`t.notThrows()`](https://github.com/sindresorhus/ava#notthrowsfunctionpromise-message)
-- [`t.pass()`](https://github.com/sindresorhus/ava#passmessage)
-- [`t.fail()`](https://github.com/sindresorhus/ava#failmessage)
+- [`t.true()`](https://github.com/avajs/ava#truevalue-message)
+- [`t.deepEqual()`](https://github.com/avajs/ava#deepequalvalue-expected-message)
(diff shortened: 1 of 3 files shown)


jamestalmage:
Hmm. Maybe not `false` / `falsy`?

Maybe not even `truthy`?

The whole point of preferring power-assert is to move all the logic inside the expression that power-assert will evaluate.

jamestalmage:
@twada? :arrow_up:

twada:
@bdougherty Thank you for updating!

> Hmm. Maybe not false / falsy?

Yeah. No double negatives. No `false`, `falsy`.

> Maybe not even truthy?

I feel it's a bit too strict (but interesting to go).
What do you think? @jamestalmage @sindresorhus

jamestalmage:
> I feel it's a bit too strict ...

Yeah, ok, let's leave it as allowed.

---

Perhaps let's take the opportunity to add something to the docs that explains the rational a bit better than we do currently?

@bdougherty - see the power-assert repo for some language. (i.e. "best API is one you don't have to remember"), etc.

sindresorhus:
I would drop `truthy` too. Imagine you have a function `foo()` which normally returns true, but suddenly returns 1 instead. `t.truthy(foo())` would not catch that case. IMHO in tests you should always be as strict as possible. It's the subtle bugs that gets you.

jamestalmage:
Maybe AVA's `truthy`/`falsy` assertions should fail when used with `boolean`s? This isn't something we could enforce with static analysis, so it would have to be done in AVA at runtime.

I will occasionally use `truthy` when I have some method that returns an object or `null`. It's usually followed immediately by another assertion on that library:

```js
const obj = someMethod();
t.truthy(obj);
t.is(obj.property, expected);
```

The second assertion would throw a TypeError, but if it is in a callback, it gets reported as an uncaught exception, and it isn't reported to the right test.

twada:
Now I agreed with @sindresorhus. Let's make it strict enough.

sindresorhus:
> Maybe AVA's truthy/falsy assertions should fail when used with booleans?

What if an API returns `true` or a string depending on the input. Should we really be that opinionated in AVA? Wish ESLint did deep type analysis...

sindresorhus:
@bdougherty Would you be able to update the PR?

bdougherty:
@sindresorhus yeah I can update sometime this week. What's the final word? Just `true` and nothing else?

sindresorhus:
Yup, remove `truthy`, `falsy`, `false.

bdougherty:
@sindresorhus updated!

sindresorhus:
Looks great. Thanks @bdougherty :)

SimenB:
Any chance for a release of this?

SimenB:
@sindresorhus ping 😄

Issue #64, opened 2016-04-06
`no-todo-implementation` rule


jfmengels:
Disallow having a function (or simply more than 1 argument) in a `test.todo`

See https://github.com/sindresorhus/ava/issues/697

jamestalmage:
`no-todo-implementation`

PR #125, opened 2016-07-01
Make test.todo message nicer

forivall:
`test.todo` is more similar to [no-warning-comments](http://eslint.org/docs/rules/no-warning-comments) rather than `no-cb` or `no-skip`: in those cases, you want them to be forbidden, whereas todo means that it should be filled in. So, "should not be used" feels like the wrong language to be used here.

Open to discussion on wording, etc. Perhaps something more like "Implementation required for `test.todo()`", since the ava error that lead me to todo was ```TypeError: Expected an implementation. Use `test.todo()` for tests without an implementation.```. Bikeshed away.

### DIFF
--- a/rules/no-todo-test.js
+++ b/rules/no-todo-test.js
@@ -12,7 +12,7 @@ module.exports = function (context) {
 			if (ava.hasTestModifier('todo')) {
 				context.report({
 					node: node,
-					message: '`test.todo()` should be not be used.'
+					message: 'Unexpected `test.todo()` stub.'
 				});
 			}
 		})


sindresorhus:
I think the intention of this rule was to be exactly like `no-cb-test` and `no-skip-test`. Some companies might not want to allow using `todo` tests as they might have a different process for handling cases like that.

> Perhaps something more like "Implementation required for test.todo()"

That's the opposite of how it works though. `test.todo` is not allowed to have an implementation.

forivall:
Well then, "Test implementation required instead of `test.todo()`", since that's what I meant by "for", but I can see who it's ambiguous.

Otherwise, I might just end up making an eslint translation pack that's friendly and uses emoji and such. :wink:

sindresorhus:
Seems like what you're after is: https://github.com/avajs/eslint-plugin-ava/issues/64

jfmengels:
I agree with @sindresorhus. This rule was meant to forbid `test.todo()`, so that you could for instance write down all your specs in your tests, and not forget leaving some unfilled before making your pull request.

The message could include something like `Please write the test for this test.todo()` IMO. No harm in telling the user what we expect him to do.

sindresorhus:
> The message could include something like Please write the test for this test.todo() IMO. No harm in telling the user what we expect him to do.

Do we really want to enforce only that use-case for this rule? As I mentioned earlier, some companies might want to forbid the rule for other reasons. I think #64 is the solution here, not changing this rule.

jfmengels:
> I think #64 is the solution here

I thought of it as a companion to #64, that one definitely needs to be added anyway.

> As I mentioned earlier, some companies might want to forbid the rule for other reasons

I have trouble finding other use cases, but they could exist, you're right.

sindresorhus:
Closing in favor of #64.

PR #130, opened 2016-07-06
Upgrade to Eslint 3 (fixes #128)

### LINKED ISSUE #128: Upgrade to ESLint 3
New rule format: http://eslint.org/blog/2016/07/eslint-new-rule-format (We should use the automatic translation)

And bump the peerDependencies and devDependencies.

jfmengels:
Upgrade to Eslint 3 (fixes #128)

I didn't use the migration tool in the end as it adds spaces everywhere and destroys the indentation we had. It also indented the main rule function which was not needed.

I think that we could add a `recommended` setting and generate the recommended settings that way. Would remove a lot of boilerplate in addition to what has been done using `req-all`. I'll leave that for later.

Let me know if I've missed anything.

### DIFF
--- a/rules/assertion-arguments.js
+++ b/rules/assertion-arguments.js
@@ -88,7 +88,7 @@ function nbArguments(node) {
 	return false;
 }
 
-module.exports = context => {
+const create = context => {
 	const ava = createAvaRule();
 	const options = context.options[0] || {};
 	const enforcesMessage = Boolean(options.message);
@@ -137,14 +137,20 @@ module.exports = context => {
 	});
 };
 
-module.exports.schema = [{
-	type: 'object',
-	properties: {
-		message: {
-			enum: [
-				'always',
-				'never'
-			]
-		}
+module.exports = {
+	create,
+	meta: {
+		docs: {},
+		schema: [{
+			type: 'object',
+			properties: {
+				message: {
+					enum: [
+						'always',
+						'never'
+					]
+				}
+			}
+		}]
 	}
-}];
+};
--- a/rules/assertion-arguments.js
+++ b/rules/assertion-arguments.js
@@ -137,20 +137,21 @@ const create = context => {
 	});
 };
 
+const schema = [{
+	type: 'object',
+	properties: {
+		message: {
+			enum: [
+				'always',
+				'never'
+			]
+		}
+	}
+}];
+
 module.exports = {
 	create,
 	meta: {
-		docs: {},
-		schema: [{
-			type: 'object',
-			properties: {
-				message: {
-					enum: [
-						'always',
-						'never'
-					]
-				}
-			}
-		}]
+		schema
 	}
 };
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
   "devDependencies": {
     "ava": "*",
     "coveralls": "^2.11.9",
-    "eslint": "^2.2.0",
+    "eslint": "^3.0.1",
     "eslint-ava-rule-tester": "^0.1.1",
     "js-combinatorics": "^0.5.0",
     "nyc": "^6.4.0",
@@ -82,7 +82,7 @@
     "xo": "*"
   },
   "peerDependencies": {
-    "eslint": ">=2"
+    "eslint": ">=3"
   },
   "xo": {
     "esnext": true,
--- a/rules/assertion-message.js
+++ b/rules/assertion-message.js
@@ -74,15 +74,17 @@ const create = context => {
 	});
 };
 
+const schema = [{
+	enum: [
+		'always',
+		'never'
+	]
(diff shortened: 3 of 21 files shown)


rules/assertion-arguments.js
```diff
@@ -137,14 +137,20 @@ module.exports = context => {
...
-			enum: [
-				'always',
-				'never'
-			]
-		}
+module.exports = {
+	create,
+	meta: {
+		docs: {},
+		schema: [{
```
sindresorhus: Can you move the `schema` into a top-level variable too?

rules/assertion-arguments.js
```diff
@@ -137,14 +137,20 @@ module.exports = context => {
...
-		message: {
-			enum: [
-				'always',
-				'never'
-			]
-		}
+module.exports = {
+	create,
+	meta: {
+		docs: {},
```
sindresorhus: Do we have to fill out this? Would prefer to leave it out if optional.

jfmengels: I'll take a look, but this is where core rules have their `recommended` setting from what I can see, so it might be a good spot to put that value. If we go that, there's no value in removing it.

rules/assertion-arguments.js
jfmengels: Sure

rules/assertion-arguments.js
jfmengels: Looks like it's optional

sindresorhus:
@jfmengels Can you remove the `docs` stuff?

jfmengels:
Done. I left the `meta` field though, even when it's empty. Again, not sure it will work without, though the tests without it.

FYI, I changed one of my plugins to generate the recommended settings from the docs field.
https://github.com/jfmengels/eslint-plugin-lodash-fp/commit/97a04abc91b6447b2823f5df4fa9b23b4e97904f

Actually, I'm not sure why the recommended setting is in the docs, and not directly in the meta, but that's how ESLint core rules are done. I'll have to look at how this is all supposed to be, and what each one is for.

sindresorhus:
We should probably mark rules as fixable in the `meta` object too:

> fixable property, which is present only if a rule is fixable (**knowing that a rule is fixable ahead of time allows us to optimize the autofixing process** as well as call this out in documentation - http://eslint.org/blog/2016/07/eslint-new-rule-format

sindresorhus:
> Actually, I'm not sure why the recommended setting is in the docs, and not directly in the meta, but that's how ESLint core rules are done. I'll have to look at how this is all supposed to be, and what each one is for.

That's weird. Do they still define a top-level recommended object like we currently do or it that up to each rule now?

jfmengels:
> We should probably mark rules as fixable in the meta object too

Agreed, but we don't have fixable rules at the moment.

I get the feeling a lot of the documentation is oriented towards the way things are done in the core rules. And they have simply set up things to look at `meta.docs` fields.

> Do they still define a top-level recommended object like we currently do or it that up to each rule now?

Up to each rule, and it's aggregated at release time, last time I checked. Don't see the point of marking stuff as recommended otherwise.

sindresorhus:
> Up to each rule, and it's aggregated at release time, last time I checked. Don't see the point of marking stuff as recommended otherwise.

Ok, so shouldn't we move https://github.com/avajs/eslint-plugin-ava/blob/fe93ad7e604ca464e6bf64a0bce96a31e9487f74/index.js#L16-L35 into each rule?

jfmengels:
:+1: The same way I did in lodash-fp?

I'm thinking of adding an anchor tag in the README, and injecting the generated recommended config in there (https://github.com/avajs/eslint-plugin-ava#usage), as it's something that we have to add manually.

Same thing for [this section](https://github.com/avajs/eslint-plugin-ava#rules), which should be even easier to do.

I'll try it in lodash-fp or some other plugin of mine, and if it pans out well I'll give a shout here.

sindresorhus:
Why wouldn't ESLint just aggregate the recommended stuff at runtime though? Feels hacky having to manually do it at publish time...

sindresorhus:
Anyways. This PR is good to go. Thanks 👍

jfmengels:
You're welcome :)

> Why wouldn't ESLint just aggregate the recommended stuff at runtime though? Feels hacky having to manually do it at publish time...

Makes sense to me.
Maybe you don't want a recommended setting, maybe you want several. Maybe you want to include core rules in your [recommended setting](https://github.com/jfmengels/eslint-plugin-fp/blob/master/index.js#L21).
It's not flexible enough IMO if ESLint tried to fetch a config automatically for a plugin using the plugin's rules' meta data.

That said, it could be an interesting idea when there is no recommended config set up. Might be worth proposing.

PR #135, opened 2016-08-07
Add `no-nested-tests` rule (fixes #131)

### LINKED ISSUE #131: `no-nested-tests` rule
Forbid the use of `test()` inside of another `test()` call.
Linked to https://github.com/avajs/ava/issues/948

jfmengels:
Add `no-nested-tests` rule (fixes #131)

### DIFF
--- a/create-ava-rule.js
+++ b/create-ava-rule.js
@@ -82,11 +82,9 @@ module.exports = () => {
 			}
 		},
 		CallExpression: node => {
-			if (!currentTestNode) {
-				if (isTestFunctionCall(node.callee)) {
-					// entering test function
-					currentTestNode = node;
-				}
+			if (isTestFunctionCall(node.callee)) {
+				// entering test function
+				currentTestNode = node;
 			}
 		},
 		'CallExpression:exit': node => {
--- a/docs/rules/no-nested-tests.md
+++ b/docs/rules/no-nested-tests.md
@@ -1,0 +1,38 @@
+# Ensure no tests are nested
+
+Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/related/eslint-plugin-ava/docs/rules/no-nested-tests.md)
+
+In AVA, you can not nest tests, i.e. create tests inside of other tests. Doing so will lead to odd behavior.
+
+
+## Fail
+
+```js
+import test from 'ava';
+
+test('foo', t => {
+	const result = foo();
+	t.true(result.foo === 'foo');
+
+	test('bar', t => {
+		t.true(result.bar === 'bar');
+	});
+});
+```
+
+
+## Pass
+
+```js
+import test from 'ava';
+
+test('foo', t => {
+	const result = foo();
+	t.true(result.foo === 'foo');
+});
+
+test('bar', t => {
+	const result = foo();
+	t.true(result.bar === 'bar');
+});
+```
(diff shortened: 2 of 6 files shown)


docs/rules/no-nested-tests.md
```diff
@@ -0,0 +1,38 @@
+# Ensure no tests are nested
+
+Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/related/eslint-plugin-ava/docs/rules/no-nested-tests.md)
+
+In AVA, you can not nest tests, i.e. create tests inside of other tests. Doing so will lead to odd behavior.
```
sindresorhus: can not => cannot

`i.e.` => `for example,`

docs/rules/no-nested-tests.md
```diff
@@ -0,0 +1,38 @@
...
+
+
+## Fail
+
+```js
+import test from 'ava';
+
+test('foo', t => {
+	const result = foo();
+	t.true(result.foo === 'foo');
```
sindresorhus: Drop the ` === 'foo'` part

docs/rules/no-nested-tests.md
```diff
@@ -0,0 +1,38 @@
...
+
+	test('bar', t => {
+		t.true(result.bar === 'bar');
```
sindresorhus: Same here.

create-ava-rule.js
```diff
@@ -82,11 +82,9 @@ module.exports = () => {
 			}
 		},
 		CallExpression: node => {
-			if (!currentTestNode) {
```
sindresorhus: Why was this removed? Will it have any side-effects for any of the other rules?

docs/rules/no-nested-tests.md
```diff
@@ -0,0 +1,38 @@
+# Ensure no tests are nested
+
+Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/related/eslint-plugin-ava/docs/rules/no-nested-tests.md)
```
sindresorhus: I assume this will be added after these rules have landed?

create-ava-rule.js
jfmengels: It was a safe-guard (or rather optimization?) to prevent entering a new test when you were already in one.

It will only be able to make a difference for other rules (but probably won't) if you have nested tests. I can improve it to make sure the other rules won't be affected, but I'm not sure that it's worth it as there is a hard stance against nested tests. Let me know what you prefer.

docs/rules/no-nested-tests.md
jfmengels: Yeah, I assumed that @forresst would work on it, as he has always been very quick on adding this. I just thought I'd make his job easier :sweat_smile:.

sindresorhus: 👌

create-ava-rule.js
sindresorhus: It's fine like this. Was just curious about the reason.

sindresorhus:
Thanks for all the hard work on these new rules. Can't believe we're at 23 rules already. 🤗

giphy

jfmengels:
chris-pratt-youre-welcome-3oEduGRYSthHoPSTGU

docs/rules/no-nested-tests.md
forresst: @sindresorhus @jfmengels Done : avajs/ava-docs@e779c852f2e0d81df0d485ce6481c63d7152b916 :smile_cat:

jfmengels: Man... Even quicker than I thought. Awesome! :D

PR #136, opened 2016-08-10
update use-t-well rule to include error method as that is an alias for ifError

bojand:


sindresorhus:
It's an alias for legacy reasons, but not something we want to recommend. It will go away eventually.

bojand:
OK good to know as I was using it. Thanks.

sindresorhus:
Noted. I'll ensure we officially deprecate it.

PR #138, opened 2016-08-29
Add `no-duplicate-modifiers` rule (fixes #137)

### LINKED ISSUE #137: `no-duplicate-modifiers` rule
Suggested in https://github.com/avajs/ava/issues/1005.

jfmengels:
Add `no-duplicate-modifiers` rule (fixes #137)

### DIFF
--- a/docs/rules/no-duplicate-modifiers.md
+++ b/docs/rules/no-duplicate-modifiers.md
@@ -1,0 +1,30 @@
+# Ensure tests do not have duplicate modifiers
+
+Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/related/eslint-plugin-ava/docs/rules/no-duplicate-modifiers.md)
+
+Prevent the use of duplicate test modifiers.
+
+
+## Fail
+
+```js
+import test = from 'ava';
+
+test.only.only(t => {});
+test.serial.serial(t => {});
+test.cb.cb(t => {});
+test.beforeEach.beforeEach(t => {});
+test.only.only.cb(t => {});
+```
+
+
+## Pass
+
+```js
+import test = from 'ava';
+
+test.only(t => {});
+test.serial(t => {});
+test.cb.only(t => {});
+test.beforeEach(t => {});
+```
--- a/docs/rules/no-duplicate-modifiers.md
+++ b/docs/rules/no-duplicate-modifiers.md
@@ -2,13 +2,13 @@
 
 Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/related/eslint-plugin-ava/docs/rules/no-duplicate-modifiers.md)
 
-Prevent the use of duplicate test modifiers.
+Prevent the use of duplicate [test modifiers](https://github.com/avajs/ava#api).
 
 
 ## Fail
 
 ```js
-import test = from 'ava';
+import test from 'ava';
 
 test.only.only(t => {});
 test.serial.serial(t => {});
@@ -21,7 +21,7 @@ test.only.only.cb(t => {});
 ## Pass
 
 ```js
-import test = from 'ava';
+import test from 'ava';
 
 test.only(t => {});
 test.serial(t => {});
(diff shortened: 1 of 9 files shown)


docs/rules/no-duplicate-modifiers.md
```diff
@@ -0,0 +1,30 @@
...
+
+Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/related/eslint-plugin-ava/docs/rules/no-duplicate-modifiers.md)
+
+Prevent the use of duplicate test modifiers.
+
+
+## Fail
+
+```js
+import test = from 'ava';
```
sindresorhus: Less `=` ;)

test/no-duplicate-modifiers.js
```diff
@@ -0,0 +1,58 @@
...
+import rule from '../rules/no-duplicate-modifiers';
+
+const ruleTester = avaRuleTester(test, {
+	env: {
+		es6: true
+	}
+});
+
+const ruleError = {ruleId: 'no-duplicate-modifiers'};
+const header = `const test = require('ava');\n`;
```
sindresorhus: Missing `)`

util.js
```diff
@@ -40,3 +40,15 @@ exports.getAvaConfig = filepath => {
 exports.isFunctionExpression = node => {
 	return node && functionExpressions.indexOf(node.type) !== -1;
 };
+
+exports.getTestModifiers = function getTestModifiers(node) {
```
sindresorhus: Arrow function

docs/rules/no-duplicate-modifiers.md
```diff
@@ -0,0 +1,30 @@
+# Ensure tests do not have duplicate modifiers
+
+Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/related/eslint-plugin-ava/docs/rules/no-duplicate-modifiers.md)
+
+Prevent the use of duplicate test modifiers.
```
sindresorhus: Linkify `test modifiers` to the relevant AVA docs section.

test/no-duplicate-modifiers.js
jfmengels: where ? :confused:

util.js
jfmengels: Doesn't work, as getTestModifiers is called recursively. It needs to either be declared before and then re-assigned to exports, or declared as a function expression, which I prefer in this case.

test/no-duplicate-modifiers.js
sindresorhus: Wut. It looked like this to me yesterday:

```js
const header = `const test = require('ava';\n`;
```

Maybe a GitHub bug.

sindresorhus: I still see it when viewing the full diff:

screen shot 2016-08-30 at 20 50 10

util.js
sindresorhus: I was thinking:

```js
exports.getTestModifiers = node => {
	if (node.type === 'CallExpression') {
		return exports.getTestModifiers(node.callee);
	}

	if (node.type === 'MemberExpression') {
		return exports.getTestModifiers(node.object).concat(node.property.name);
	}

	return [];
};
```

But you're right, doesn't really matter.

test/no-duplicate-modifiers.js
jfmengels: Huh weird. Definitely a GitHub thing, I see it in my editor (and pretty sure no tests would have passed otherwise)

util.js
jfmengels: I really don't like that way of calling functions, as you're mixing how to call functions with how/whether they are exported, even though it's easy to disconnect the two.
I see it all the time at work, and it's frustrating every time :p

sindresorhus: Good point. Will be moot anyways when we get ES modules in Node.js.

jfmengels: Looking forward to it, but no idea how long that'll take before we can put that in modules without transpiling... End of life of Node v6 maybe...? :/

sindresorhus: Probably EOL Node.js 8...

sindresorhus:
Geweldig!

util.js
jfmengels: So April 2019 or 2020... yay...

PR #143, opened 2016-09-24
Do not count context methods as assertion methods (fixes #142)

### LINKED ISSUE #142: max-asserts incorrectly counts called functions on t.context as assertions
### Problem
`ava/max-asserts` incorrectly counts called functions on `t.context` as assertions.

### Example:
```js
/* eslint "ava/max-asserts" : [ "warn", 5 ] */

import test from 'ava';

test.beforeEach( t => {
	t.context.bar = () => 'bar';
});


// this will be fine
test( 'max-asserts ok', t => {
	const foo = 'foo';

	t.true( typeof foo === 'string' );
	t.true( typeof foo === 'string' );
	t.true( typeof foo === 'string' );
	t.true( typeof foo === 'string' );
	t.true( typeof foo === 'string' );
});


// this will count 6 assertions
test( 'max-asserts incorrect count', t => {
	const foo = 'foo';

	// this is counted as an assertion
	const bar1 = t.context.bar();

	// this is not counted as an assertion
	const Bar = t.context.bar;
	const bar2 = Bar();

	t.true( typeof foo === 'string' );
	t.true( typeof foo === 'string' );
	t.true( typeof foo === 'string' );
	t.true( typeof foo === 'string' );
	t.true( typeof foo === 'string' );
});
```

### Expected
No warnings

### Actual
The following warnings are displayed when linting the above file:
```
37:2  warning  Expected at most 5 assertions, but found 6  ava/max-asserts
✖ 1 problem (0 errors, 1 warning)
```

jfmengels:
Do not count context methods as assertion methods (fixes #142)

### DIFF
--- a/rules/use-t-well.js
+++ b/rules/use-t-well.js
@@ -3,9 +3,7 @@ const visitIf = require('enhance-visitors').visitIf;
 const util = require('../util');
 const createAvaRule = require('../create-ava-rule');
 
-const methods = util.assertionMethods.concat(['end', 'plan']);
-
-const isMethod = name => methods.indexOf(name) !== -1;
+const isMethod = name => util.executionMethods.has(name);
 
 const isCallExpression = node =>
 	node.parent.type === 'CallExpression' &&
--- a/rules/use-t-well.js
+++ b/rules/use-t-well.js
@@ -3,37 +3,9 @@ const visitIf = require('enhance-visitors').visitIf;
 const util = require('../util');
 const createAvaRule = require('../create-ava-rule');
 
-const methods = [
-	'end',
-	'pass',
-	'fail',
-	'truthy',
-	'falsy',
-	'true',
-	'false',
-	'is',
-	'not',
-	'deepEqual',
-	'notDeepEqual',
-	'throws',
-	'notThrows',
-	'regex',
-	'notRegex',
-	'ifError',
-	'plan'
-];
+const methods = util.assertionMethods.concat(['end', 'plan']);
 
 const isMethod = name => methods.indexOf(name) !== -1;
-
-const getMembers = node => {
-	const name = node.property.name;
-
-	if (node.object.type === 'MemberExpression') {
-		return getMembers(node.object).concat(name);
-	}
-
-	return [name];
-};
 
 const isCallExpression = node =>
 	node.parent.type === 'CallExpression' &&
@@ -84,7 +56,7 @@ const create = context => {
 				return;
 			}
 
-			const members = getMembers(node);
+			const members = util.getMembers(node);
 			const stats = getMemberStats(members);
 
 			if (members[0] === 'context') {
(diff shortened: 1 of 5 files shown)


util.js
```diff
@@ -52,3 +52,36 @@ exports.getTestModifiers = function getTestModifiers(node) {
...
+	if (node.object.type === 'MemberExpression') {
+		return getMembers(node.object).concat(name);
+	}
+
+	return [name];
+};
+
+exports.getMembers = getMembers;
+
+const assertionMethodsNbArguments = {
```
sindresorhus: I would name it: `assertionMethodNumArgs`. Num is a more common short name for Number.

rules/use-t-well.js
```diff
@@ -3,38 +3,10 @@ const visitIf = require('enhance-visitors').visitIf;
...
-	'deepEqual',
-	'notDeepEqual',
-	'throws',
-	'notThrows',
-	'regex',
-	'notRegex',
-	'ifError',
-	'plan'
-];
+const methods = util.assertionMethods.concat(['end', 'plan']);
```
sindresorhus: I think this should be in `util.js` too. Name it `util.contextMethods`.

util.js
sindresorhus: I think we should use a `Map` here instead.

util.js
```diff
@@ -52,3 +52,36 @@ exports.getTestModifiers = function getTestModifiers(node) {
...
+	pass: 0,
+	regex: 2,
+	notRegex: 2,
+	throws: 1,
+	true: 1,
+	truthy: 1
+};
+
+exports.assertionMethodsNbArguments = assertionMethodsNbArguments;
+exports.assertionMethods = Object.keys(assertionMethodsNbArguments);
```
sindresorhus: And this should be a `Set`.

rules/use-t-well.js
jfmengels: contextMethods? They're not part of `t.context` :confused:

util.js
jfmengels: Why the preference for Map and Set? Except for holding objects as keys for Map, I don't know the benefits. Don't mind changing it, as I don't see drawbacks to it, mostly curious.

jfmengels: Changed to Map/Set, but would still like your opinion on its benefits :)

rules/use-t-well.js
jfmengels: Named it executionMethods, as it's called the execution object in the docs.

util.js
sindresorhus: They're more proper for data, better API, faster, and can handle any kind of key name.

rules/use-t-well.js
sindresorhus: Ah right. We used to call `t` the test context before we got `t.context`.

jfmengels:
Aw man... forgot `...` wasn't supported in v4... -_-'

sindresorhus:
Looks great! :)

PR #148, opened 2016-10-11
Fix no-async-fn-without-await when used with default parser (fixes #147)

### LINKED ISSUE #147: no-async-fn-without-await errors when await is last call
```js
test.after(async () => {
  await app.stop();
});
```

Is this because the test could just return the promise explicitly instead? If so maybe it would be better to split it into its own rule, it's kind of confusing at the moment.

jfmengels:
Fix no-async-fn-without-await when used with default parser (fixes #147)

Espree supports async/await, but it doesn't generate the same AST as babel-eslint when using async/await. Espree uses a `AwaitExpression` node (which makes ore sense IMO), while `babel-eslint` uses `YieldExpression` (which I think could give incorrect results when using the generator `yield` keyword).

The test file looks like it's been rewritten, but I just indented the whole test suite, to be able to use it with both parsers (I've added one test, and that's it).

The tests fail because of XO, I'll do that in another PR real quick.
EDIT: Done in #149. Tests for this PR should be re-run after it gets merged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "ava": "*",
     "babel-eslint": "^6.1.2",
     "coveralls": "^2.11.9",
-    "eslint": "^3.0.1",
+    "eslint": "^3.7.1",
     "eslint-ava-rule-tester": "^2.0.0",
     "js-combinatorics": "^0.5.0",
     "nyc": "^7.1.0",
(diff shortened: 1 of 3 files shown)


sindresorhus:
LGTM when merge conflict is fixed. Thanks for fixing this so fast @jfmengels 💃

jfmengels:
No problem ;)

PR #149, opened 2016-10-11
Fix linting errors from XO 0.17

jfmengels:
Fix linting errors from XO 0.17

### DIFF
--- a/rules/no-statement-after-end.js
+++ b/rules/no-statement-after-end.js
@@ -51,7 +51,7 @@ const create = context => {
 			.filter(info => info.ended);
 
 		// If this segment or any previous segment is already ended, further statements are not allowed, report as an error.
-		if (ended.length) {
+		if (ended.length !== 0) {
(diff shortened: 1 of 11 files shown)


rules/no-statement-after-end.js
```diff
@@ -51,7 +51,7 @@ const create = context => {
 			.filter(info => info.ended);
 
 		// If this segment or any previous segment is already ended, further statements are not allowed, report as an error.
-		if (ended.length) {
+		if (ended.length !== 0) {
```
sindresorhus: Would be more natural here to use `> 0`.

sindresorhus: I've changed it on merge.

jfmengels: Sure, fine by me.

If this nitpick happens a lot for you, then it could probably deserve a linting rule, or options for the explicit-length-check rule. We could have another linting rule for comparing the result of `.indexOf`.

PR #153, opened 2016-10-27
Set `ecmaVersion` to `2017`

SimenB:
Probably a reason you haven't done it yet (other than the fact it's a breaking change), but it's annoying having ava override our setting just because we want to use recommended config, which means we have to specify `ecmaVersion` multiple places (as we only have ava in test dirs)

An idea might be to drop the version entirely from recommended? That's still breaking, though.

Could do `require('eslint/package.json').version` and check if it's new enough to keep it non-breaking?

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,7 +9,7 @@ module.exports = {
 				es6: true
 			},
 			parserOptions: {
-				ecmaVersion: 7,
+				ecmaVersion: 2017,
 				sourceType: 'module'
 			},
 			rules: {
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "xo": "*"
   },
   "peerDependencies": {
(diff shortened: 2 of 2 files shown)


sindresorhus:
> An idea might be to drop the version entirely from recommended? That's still breaking, though.

I think the reason we set it is that AVA uses Babel, so you can use the latest syntax without setting anything. We should always have this to the latest mode ESLint provides.

We've already planned a breaking release, so can include this then.

SimenB:
> We should always have this to the latest mode ESLint provides.

Would be cool if they accepted `latest` or something, kinda like https://babeljs.io/docs/plugins/preset-latest/

sindresorhus:
@SimenB Indeed. Open an issue on ESLint ;)

PR #161, opened 2016-11-27
Fix `no-async-without-fn` when using a title

### LINKED ISSUE #160: Rule `no-async-fn-without-await` doesn't report when test has a title
The following reports:

```js
test(async t => {});
```

But this does not:

```js
test('foo', async t => {});
```

---

We should audit our tests for all our rules to ensure they all have a test for with and without a title.

---

// @jfmengels

jfmengels:
Fixes #160: Fixes `no-async-without-fn` not reporting anything when not using a title.

I decided to simply check if any of the first two arguments is a async. I thought about writing a function that finds the implementation function among the arguments, but that can be rather tricky when using identifiers.

```js
function implementationFn(t, argFunction) {
  t.throws(argFunction(1));
}
implementationFn.title = 'foo'
test(implementationFn, async n => {}); // We'd probably assume the second function is the implementation function, but no.
```

This could create false positives in the case shown above, if that function also does not contain `async`. Sounds like an extreme edge case, and they should not mark it as async anyway IMO, so it should be fine.

### DIFF
--- a/rules/no-async-fn-without-await.js
+++ b/rules/no-async-fn-without-await.js
@@ -13,13 +13,14 @@ const create = context => {
 		}
 	};
 
+	const isAsync = node => Boolean(node && node.async);
+
 	return ava.merge({
 		CallExpression: visitIf([
 			ava.isInTestFile,
 			ava.isTestNode
 		])(node => {
-			const implementationFn = node.arguments[0];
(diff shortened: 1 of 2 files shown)


sindresorhus:
Good fix :)

I agree your example is an extreme edge-case we shouldn't care about.

PR #165, opened 2017-01-09
Allow passing an argument to `t.end()` (fixes #164)

### LINKED ISSUE #164: ava/assertion-arguments argues about explicit call of t.end() in test.cb()
First of all i am not sure if this is a bug report or a feature request. I ended up deactivating the `assertion-arguments`-rule.

Given the following condensed test-code example, where i want to test if the method takes path `b`:

```js
test.cb(t => {
  m({
    a: {
      callback: () => t.end(true)
    },
    b: {
      callback: () => t.end(false)
    }
  });
});
```

`ava/assertion-arguments` complains: Too many arguments. Expected at most 0.

I know the documentation somehow exposes `t.end()` as single callback handler without any parameters. But it also says if "any truthy value [will be] passed as the first argument to t.end" it treats it as error.

@jfmengels already mentioned using t.fail/t.success instead. Unfortunately this is prohibited by `ava/test-ended`: Callback test was not ended. Make sure to explicitly end the test with t.end().

It's probably just a decision which rule to adjust, however. If there's something i can do. Let me know.

jfmengels:
Allow passing an argument to `t.end()` (fixes #164)

`t.end()` is now a special case for this rule, as the optional argument it allows is not a message that we'd like to enforce based on the `message` option.

I've simplified the implementation a bit by re-using `util.getMembers`.

### DIFF
--- a/rules/assertion-arguments.js
+++ b/rules/assertion-arguments.js
@@ -8,10 +8,6 @@ const expectedNbArguments = {
 		min: 2,
 		max: 3
 	},
-	end: {
-		min: 0,
-		max: 0
-	},
 	fail: {
 		min: 0,
 		max: 1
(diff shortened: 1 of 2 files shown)


sindresorhus:
Looks great!

leo-yay

Issue #21, opened 2016-02-27
Autofix rules


sindresorhus:
Would be useful to support autofixing for some of the rules. Not all are possible.

http://eslint.org/docs/developer-guide/working-with-rules#contextreport
- [ ] `no-skip-test` (just remove the `skip` modifier)
- [x] `no-only-test` (just remove the `only` modifier)
- [ ] `no-identical-title` (append #1, #2, etc, to the title)
- [ ] `no-skip-assert` (just remove the `skip` modifier)

Can't think of any way to reliably fix the other rules, but input welcome!

jfmengels:
:+1: Great idea.

Not sure about `no-identical-title` though, as they won't really make things more readable in the source code (only when matching AVA's output to a test). Don't forget that the rule can also be triggered because there is no title.

sindresorhus:
> Not sure about no-identical-title though, as they won't really make things more readable in the source code (only when matching AVA's output to a test).

Appending a `#n` does make it clear it's testing the same thing just split up for readability, but I don't feel strongly about it, and is a pretty edge-case anyways.

jamestalmage:
W/regard to the `no-skip-*` rules, I think auto fixing those is problematic. It will likely make the tests fail. If I am running `xo --fix` to clear up some whitespace / semicolon mistakes - I probably don't want you messing with my tests/assertions. 

I don't think we should auto-fix anything that modifies code behavior (unless maybe the user sets some other CLI option or environment variable? `FIX_AVA=true xo --fix`. 

One exception to the above. I think removing `only` is fine. Users should never check in `.only` tests, but `.skip` is acceptable in a number of situations (i.e. somebody has submitted a PR with a failing test but not the fix itself).

sindresorhus:
But without it will make XO fail. So it will still fail.

jamestalmage:
Does `--fix` only fix errors, or warnings too?

sindresorhus:
Warnings too.

florianb:
@sindresorhus: i would like to give it a try if it is still relevant.

sindresorhus:
@florianb Sure. Go ahead. Seems like we couldn't agree on all of these yet, but you could start with autofixing for `no-only-test` :)

florianb:
Great - i am sure a PR will stimulate discussions.. 😄

florianb:
[...]

PR #173, opened 2017-03-31
Add fixer for the `no-only-test` rule

florianb:


rules/no-only-test.js
```diff
@@ -13,7 +27,12 @@ const create = context => {
 			if (ava.hasTestModifier('only')) {
 				context.report({
 					node,
-					message: '`test.only()` should not be used.'
+					message: '`test.only()` should not be used.',
+					fix: fixer => {
+						const range = getTestModifier(node, 'only').range;
+						range[0] -= 1;
+						return fixer.remove(getTestModifier(node, 'only'));
```
sindresorhus: I would do:

```js
const range = getTestModifier(node, 'only').range.slice();
range[0] -= 1;
return fixer.removeRange(range);
```

rules/no-only-test.js
```diff
@@ -2,6 +2,20 @@
 const visitIf = require('enhance-visitors').visitIf;
 const createAvaRule = require('../create-ava-rule');
 
+const getTestModifier = function (node, mod) {
```
sindresorhus: Use arrow-function.

sindresorhus: And move it to `util.js`

rules/no-only-test.js
```diff
@@ -2,6 +2,20 @@
...
+		return getTestModifier(node.callee, mod);
+	} else if (node.type === 'MemberExpression') {
+		if (node.property.type === 'Identifier' && node.property.name === mod) {
+			return node.property;
+		}
+
+		return getTestModifier(node.object, mod);
+	}
+
+	return undefined;
```
sindresorhus: This is not needed.

sindresorhus:
Can you also mark it `fixable` in the docs, like done here https://github.com/sindresorhus/eslint-plugin-unicorn#rules

rules/no-only-test.js
florianb: ✔️

rules/no-only-test.js
florianb: ✔️

florianb:
Okay @sindresorhus - i finally made the changes, apologies for the delay.

Moving `getTestModifier` to `util` made it necessary using a named function instead of an arrow function. Let me know if further changes are needed. 🍸

jfmengels:
Hi @florianb, thanks for your work on this :)

I would like to add three cases to the tests. The third one fails and currently creates incorrect JS code (`test.(t => { t.pass(); });`).

```js
		{
			code: header + 'test\n  .only(t => { t.pass(); });',
			output: header + 'test\n  (t => { t.pass(); });',
			errors
		},
		{
			code: header + 'test  .only(t => { t.pass(); });',
			output: header + 'test  (t => { t.pass(); });',
			errors
		},
		{
			code: header + 'test.\n  only(t => { t.pass(); });',
			output: header + 'test\n  (t => { t.pass(); });',
			errors
		}
```

That said, I'm actually against having this rule be fixable :thinking: 
I use autofix a lot now (manually), for things like formatting (especially since I started using prettier and a lot of core rules became autofixable), and some people even have it run automatically on save. Having this rule be fixable means that you can not run the autofix (or even save your file?) while you want to run some tests with `only`.

florianb:
Hi @jfmengels - thanks for suggesting the test-cases, they are perfectly valid!

I absolutely understand your opinion, to be hones, i picked that rule since i wanted to help out and the `no-only`-rule wasn't discussed controversial in #21. I will wait with any further activities until the members found a consent on this.

But to add something to the discussion - i use fixes to make my current work compliant before pushing to the remote.

sindresorhus:
> Having this rule be fixable means that you can not run the autofix (or even save your file?) while you want to run some tests with only.

That can be solved in the editor plugins though, like https://github.com/AtomLinter/linter-eslint/pull/795.

jfmengels:
Ah, I didn't know about that, that's great to hear.
I think most people will not use that, but it's probably worth trying adding it. If it bothers a lot of people, we can always explain in the docs how to disable the autofixing in the editor.
:+1: for me then, pending the addition of the tests that I put above.

sindresorhus:
And in atom-linter-xo, I can disable the auto-fix for it by default.

florianb:
Aye - thanks, going to add the test. ⏳

florianb:
@jfmengels - i added additional tests using tabs instead of spaces.
@sindresorhus - done, even though i guess you'll have to refine my style for your final pleasure. 😄

PR #181, opened 2017-11-13
add support for t.log

ykantema:
as I use AVA ,I need use the log method, and pass the eslint


sindresorhus:
Support for `t.log()` was added in https://github.com/avajs/eslint-plugin-ava/commit/033a72fd2b7e8ca064b204317d1fa04276a79574. Is that not working for you?

PR #182, opened 2017-11-17
Improve target reporting

jfmengels:
Hi all! It's been a while :sweat_smile: 

I've been increasingly growing annoyed by some error reports that I have in my IDE while developping.
One of the most annoying ones is using `test.only`, which targets the whole test call (in IDEs, it puts puts squiggly lines under the targeted code, as in the screenshot below), which makes it hard to notice other ESLint errors.

image

I started working on this and though that the other rules were easy to fix too, so I just did all those that I could think of. Let me know if you prefer that I split this into multiple PRs.

List of changes:
- no-only-test: Now targets the `only` property instead of the whole test call
- no-skip-test: Now targets the `skip` property instead of the whole test call
- no-unknown-modifiers: Now targets the unknown properties instead of the whole test call
- no-duplicate-modifiers: Now targets the duplicate properties (not the first one) instead of the whole test call
- no-identical-title: Now targets the title argument instead of the whole test call
- no-async-fn-without-await: Now targets the `await` modifier instead of the whole test call (Note: we could probably autofix this pretty easily now. What do you think?).
- no-cb-test: Now targets the `cb` property instead of the whole test call
- no-cb-test: Noticed a typo in the error message, so I changed it
  - from: '`test.cb()` should be not be used.'
  - to: '`test.cb()` should not be used.'
- I did a little bit of refactoring around the functions that dealed with getting test modifiers. There were a few ones that did close to the same thing regarding AST "exploring", so I changed it so that that part is only done at one place, and the other functions are using it.

### DIFF
--- a/create-ava-rule.js
+++ b/create-ava-rule.js
@@ -2,6 +2,7 @@
 const espurify = require('espurify');
 const enhance = require('enhance-visitors');
 const deepStrictEqual = require('deep-strict-equal');
+const util = require('./util');
 
 const avaImportDeclarationAst = {
 	type: 'ImportDeclaration',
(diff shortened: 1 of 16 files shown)


sindresorhus:
@jfmengels Nice! Looks good to me. Can you fix the merge conflict?

jfmengels:
@sindresorhus Done :+1:

sindresorhus:
Great! Really useful improvement :)

PR #184, opened 2018-01-08
Add URL to rule documentation to the metadata

Arcanemagus:
ESLint v4.15.0 added an official location for rules to store a URL to their documentation in the rule metadata in eslint/eslint#9788. This adds the URL to all the existing rules so anything consuming them can
know where their documentation is without having to resort to external packages to guess.

### DIFF
--- a/rules/assertion-message.js
+++ b/rules/assertion-message.js
@@ -68,7 +68,7 @@ module.exports = {
 	create,
 	meta: {
 		docs: {
-			url: 'https://github.com/avajs/eslint-plugin-ava/tree/master/docs/rules/assertion-message.md'
+			url: 'https://github.com/avajs/eslint-plugin-ava/blob/4211212daf1bfcfff3ebc5d4efdc4ba1a87acbf1/docs/rules/assertion-message.md'
 		},
 		schema,
 		deprecated: true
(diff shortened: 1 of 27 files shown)


rules/assertion-message.js
```diff
@@ -67,6 +67,9 @@ const schema = [{
 module.exports = {
 	create,
 	meta: {
+		docs: {
+			url: 'https://github.com/avajs/eslint-plugin-ava/tree/master/docs/rules/assertion-message.md'
```
jfmengels: This should be  https://github.com/avajs/eslint-plugin-ava/blob/4211212daf1bfcfff3ebc5d4efdc4ba1a87acbf1/docs/rules/assertion-message.md. The rule was removed after 3.1.1 and the associated file does not exist in master anymore.

Arcanemagus: Ah! A great example of why having the URL be unique to the rule is a good thing.

Arcanemagus:
Fixed @jfmengels 😉.

sindresorhus:
@Arcanemagus I know I'm late to this, but could you use a method to get the URL here too, like https://github.com/sindresorhus/eslint-plugin-unicorn/commit/875344d8d8b36ab5defa46cca8a80f3e7795fdd7#diff-01924effba2e2d51bb828e7602de3db0

Arcanemagus:
@sindresorhus Updated, note that this repo needs the `commitHash` option so I added that back to the function 😉.

sindresorhus:
Thank you :)

PR #188, opened 2018-02-15
Do not use module.parent for resolving rule name

sudo-suhas:
`module.parent` is cached after the first `require` and this results in nearly all rules having the URL to `assertion-arguments`. Here's a simple repro:

```js
'use strict';

const { rules } = require('eslint-plugin-ava');

const wrongLinks = [];

for (const ruleName of Object.keys(rules)) {
    const rule = rules[ruleName];
    const { url } = rule.meta.docs;
    if (!url.includes(ruleName)) wrongLinks.push(ruleName);
}

console.log(wrongLinks);
```

```
λ node rule-docs-url.js
[ 'max-asserts',
  'no-async-fn-without-await',
  'no-cb-test',
  'no-duplicate-modifiers',
  'no-identical-title',
  'no-ignored-test-files',
  'no-invalid-end',
  'no-nested-tests',
  'no-only-test',
  'no-skip-assert',
  'no-skip-test',
  'no-statement-after-end',
  'no-todo-implementation',
  'no-todo-test',
  'no-unknown-modifiers',
  'prefer-async-await',
  'prefer-power-assert',
  'test-ended',
  'test-title',
  'use-t-well',
  'use-t',
  'use-test',
  'use-true-false' ]
```

Credits to @SimenB for calling this out in https://github.com/jest-community/eslint-plugin-jest/commit/e0160e59e939e633282443f2fe5768ef09da2c48#r27576164.

### DIFF
--- a/rules/assertion-arguments.js
+++ b/rules/assertion-arguments.js
@@ -150,7 +150,7 @@ module.exports = {
 	create,
 	meta: {
 		docs: {
-			url: util.getDocsUrl()
+			url: util.getDocsUrl(__filename)
 		},
 		schema
 	}
--- a/rules/assertion-message.js
+++ b/rules/assertion-message.js
@@ -68,7 +68,7 @@ module.exports = {
 	create,
 	meta: {
 		docs: {
-			url: util.getDocsUrl('assertion-message', '4211212daf1bfcfff3ebc5d4efdc4ba1a87acbf1')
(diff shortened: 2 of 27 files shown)


sindresorhus:
@sudo-suhas Thanks :) Would you be able to do the same change on https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/utils/get-docs-url.js too?

sudo-suhas:
Sure, will do it by tomorrow.

PR #192, opened 2018-05-31
Add support for the ava.config.js

HenriBeck:
Adds support for checking the `ava.config.js` when there is no `ava` config inside `package.json`.
This uses the `esm` module for parsing the config file.

### DIFF
--- a/util.js
+++ b/util.js
@@ -2,6 +2,7 @@
 const fs = require('fs');
 const path = require('path');
 const pkg = require('./package');
+const esmRequire = require('esm')(module);
 
 const functionExpressions = [
 	'FunctionExpression',
@@ -33,7 +34,15 @@ exports.getAvaConfig = filepath => {
 
 	try {
 		const packageContent = JSON.parse(fs.readFileSync(filepath, 'utf8'));
-		return (packageContent && packageContent.ava) || defaultResult;
+
+		if (packageContent && packageContent.ava) {
+			return packageContent.ava;
+		}
+
+		const fileOptions = path.parse(filepath);
+		const avaConfig = esmRequire(path.join(fileOptions.dir, 'ava.config.js'));
+
+		return avaConfig.default;
 	} catch (err) {
 		return defaultResult;
 	}
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
 		"arrify": "^1.0.1",
 		"deep-strict-equal": "^0.2.0",
 		"enhance-visitors": "^1.0.0",
+		"esm": "^3.0.41",
 		"espree": "^3.1.3",
 		"espurify": "^1.5.0",
 		"import-modules": "^1.1.0",


util.js
```diff
@@ -33,7 +34,15 @@ exports.getAvaConfig = filepath => {
 
 	try {
 		const packageContent = JSON.parse(fs.readFileSync(filepath, 'utf8'));
-		return (packageContent && packageContent.ava) || defaultResult;
+
+		if (packageContent && packageContent.ava) {
```
novemberborn: Strictly speaking AVA ignores `"ava": {}` in favor of an `ava.config.js` file. We should do the same here.

util.js
```diff
@@ -33,7 +34,15 @@ exports.getAvaConfig = filepath => {
...
+			return packageContent.ava;
+		}
+
+		const fileOptions = path.parse(filepath);
+		const avaConfig = esmRequire(path.join(fileOptions.dir, 'ava.config.js'));
+
+		return avaConfig.default;
```
novemberborn: The config file may export a factory method that's invoked with an options argument. For here, that should be `{projectDir: fileOptions.dir}`.

We'll need to ignore `MODULE_NOT_FOUND` errors and return `defaultResult` instead.

util.js
HenriBeck: From the docs:

```
To use an ava.config.js file:

It must be in the same directory as your package.json
Your package.json must not contain an ava property (or, if it does, it must be an empty object)
```
I thought it would prefer the ava object from package.json

HenriBeck:
@novemberborn I added support for the function. 
About the first change request, either I miss understood the ava docs or they are wrong.

util.js
novemberborn: No it throws an exception if it finds config in `package.json` (that is, the `"ava"` key contains a non-empty object) as well as an `ava.config.js` file. Therefore if you find an empty object here, you need to check for `ava.config.js` as well. But it may not exist, and that should be OK.

I'm not sure what we should do here when both configurations are provided. If we throw an exception how does ESLint surface that?

util.js
```diff
@@ -2,6 +2,7 @@
 const fs = require('fs');
 const path = require('path');
 const pkg = require('./package');
+const esmRequire = require('esm')(module);
```
novemberborn: To be [in sync with AVA itself](https://github.com/avajs/ava/blob/b6a95b14099df88d33b12b998f41fa9bb2bbc60d/lib/load-config.js#L17:L20), you need to set `{cjs: false, mode: 'all'}` options.

util.js
HenriBeck: When you throw an error in ESLint, it just ends the process. So I think throwing an error isn't an option here.

novemberborn:
I've fixed the merge conflicts and brought the logic in line with AVA itself. I've also added some tests and updated the relevant rule documentation.

@HenriBeck are you happy with these changes?

@sindresorhus any thoughts? Would be good to merge and ship so we're compatible with both the latest AVA and ESLint.

HenriBeck:
Yeah, I think it's the right thing to return the default config when there is no config/invalid config because throwing an error would completely abort eslint.

A rule could be added which would validate the config maybe.

PR #193, opened 2018-06-01
Add `no-import-test-file` rule

### LINKED ISSUE #141: Forbid importing / requiring test files
Using the same regex logic that we use in https://github.com/avajs/eslint-plugin-ava/blob/master/docs/rules/no-ignored-test-files.md, forbid importing a test file.

There is no good reason to import a test file. If you do, you are probably:
- re-using a helper function or test macro from a different test, which should be moved in a separate helper file
- Importing a test file from a source file, which you shouldn't mix
- Naming a source file as something that will be considered by AVA as a test file

HenriBeck:
Fixes #141 

Created a rule that forbids test files in any form.

### DIFF
--- a/docs/rules/no-import-test-files.md
+++ b/docs/rules/no-import-test-files.md
@@ -1,8 +1,8 @@
-# Ensure no tests are imported anywhere
+# Ensure no test files are imported anywhere
 
 This rule will verify that you don't import any test files. It will consider the root of the project to be the closest folder containing a `package.json` file, and will not do anything if it can't find one. Test files in `node_modules` will not be linted as they are ignored by ESLint.
 
-Note that this rule will not be able to warn correctly if you use AVA by specifying the files in the command line ( `ava "lib/**/*.test.js"` ). Prefer configuring AVA as described in the link above.
+Note that this rule will not be able to warn correctly if you use AVA by specifying the files in the command line ( `ava "lib/**/*.test.js"` ). Prefer configuring AVA as described [here](https://github.com/sindresorhus/ava#configuration).
 
 ## Fail
 
@@ -10,7 +10,9 @@ Note that this rule will not be able to warn correctly if you use AVA by specify
 // File: src/index.js
 // Invalid because *.test.js is considered as a test file.
 import tests from './index.test.js';
+```
 
+```js
 // File: src/index.js
 // Invalid because any files inside __tests__ are considered test files
 import tests from './__tests__/index.js';
@@ -18,7 +20,9 @@ import tests from './__tests__/index.js';
 test('foo', t => {
 	t.pass();
 });
+```
 
+```js
 // File: utils/index.js
 // with { "files": ["lib/**/*.test.js", "utils/**/*.test.js"] }
 // in either `package.json` under 'ava key' or in the rule options
@@ -37,10 +41,14 @@ test('foo', t => {
 // File: src/index.js
 import sinon from 'sinon';
 
+```
 
+```js
 // File: src/index.js
 import utils from './utils';
+```
 
+```js
 // File: lib/index.js
 // with { "files": ["lib/**/*.test.js", "utils/**/*.test.js"] }
 // in either `package.json` under 'ava key' or in the rule options
--- a/docs/rules/no-import-test-files.md
+++ b/docs/rules/no-import-test-files.md
@@ -1,0 +1,60 @@
+# Ensure no tests are imported anywhere
+
+This rule will verify that you don't import any test files. It will consider the root of the project to be the closest folder containing a `package.json` file, and will not do anything if it can't find one. Test files in `node_modules` will not be linted as they are ignored by ESLint.
+
+Note that this rule will not be able to warn correctly if you use AVA by specifying the files in the command line ( `ava "lib/**/*.test.js"` ). Prefer configuring AVA as described in the link above.
+
+## Fail
+
+```js
+// File: src/index.js
+// Invalid because *.test.js is considered as a test file.
+import tests from './index.test.js';
+
+// File: src/index.js
+// Invalid because any files inside __tests__ are considered test files
+import tests from './__tests__/index.js';
+
+test('foo', t => {
+	t.pass();
+});
+
+// File: utils/index.js
+// with { "files": ["lib/**/*.test.js", "utils/**/*.test.js"] }
+// in either `package.json` under 'ava key' or in the rule options
+// Invalid because the imported file matches lib/**/*.test.js
+import tests from '../lib/index.test.js';
+
+test('foo', t => {
+	t.pass();
+});
+```
+
+
+## Pass
+
+```js
+// File: src/index.js
+import sinon from 'sinon';
+
+
+// File: src/index.js
+import utils from './utils';
+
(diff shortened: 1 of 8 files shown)


rules/no-import-test-files.js
```diff
@@ -0,0 +1,102 @@
...
+	'test/**/*.js',
+	'**/__tests__/**/*.js',
+	'**/*.test.js'
+];
+
+function isTestFile(files, rootDir, sourceFile, importedFile) {
+	const absoluteImportedPath = path.resolve(path.dirname(sourceFile), importedFile);
+	const relativePath = path.relative(rootDir, absoluteImportedPath);
+
+	console.log(sourceFile, importedFile, absoluteImportedPath, relativePath);
```
novemberborn: Could you remove this line?

rules/no-import-test-files.js
```diff
@@ -0,0 +1,102 @@
...
+			if (!node.arguments[0] || node.arguments[0].type !== 'Literal') {
+				return;
+			}
+
+			const isImportingTestFile = isTestFile(files, packageInfo.rootDir, filename, node.arguments[0].value);
+
+			if (isImportingTestFile) {
+				context.report({
+					node,
+					message: `You are importing a test file`,
```
novemberborn: This doesn't seem like the most helpful message, but I'm not really familiar with how this shows up in linter reports. @sindresorhus?

rules/no-import-test-files.js
```diff
@@ -0,0 +1,102 @@
...
+}
+
+const create = context => {
+	const filename = context.getFilename();
+
+	if (filename === '<text>') {
+		return {};
+	}
+
+	const packageInfo = getPackageInfo();
```
novemberborn: Could you rename this to `projectInfo` / `getProjectInfo()` etc? That better aligns with how things are named in AVA itself.

rules/no-import-test-files.js
HenriBeck: Ohh, yeah totally forgot that line.

HenriBeck:
No, this PR doesn't necessarily need #192. 

Personally, I would put that into the recommended config, because I think this is a pretty clear anti-pattern.

rules/no-import-test-files.js
HenriBeck: If you have a better commit message, feel free to suggest it.

It shows up as the message in the linter at the specific line.

sindresorhus:
> @avajs/core should this be a recommended rule?

Yes, this should be a recommended rule.

rules/no-import-test-files.js
```diff
@@ -0,0 +1,100 @@
...
+const multimatch = require('multimatch');
+const util = require('../util');
+
+const defaultFiles = [
+	'test.js',
+	'test-*.js',
+	'test/**/*.js',
+	'**/__tests__/**/*.js',
+	'**/*.test.js'
+];
```
sindresorhus: This should be moved to `util` as it's also defined in `no-ignored-test-files` (the one there should be moved too)

docs/rules/no-import-test-files.md
```diff
@@ -0,0 +1,60 @@
+# Ensure no tests are imported anywhere
```
sindresorhus: `tests` => `test files`

docs/rules/no-import-test-files.md
```diff
@@ -0,0 +1,60 @@
...
+
+This rule will verify that you don't import any test files. It will consider the root of the project to be the closest folder containing a `package.json` file, and will not do anything if it can't find one. Test files in `node_modules` will not be linted as they are ignored by ESLint.
+
```
sindresorhus: The rule should mention why it's bad to import test files, like described in https://github.com/avajs/eslint-plugin-ava/issues/141

docs/rules/no-import-test-files.md
```diff
@@ -0,0 +1,60 @@
...
+Note that this rule will not be able to warn correctly if you use AVA by specifying the files in the command line ( `ava "lib/**/*.test.js"` ). Prefer configuring AVA as described in the link above.
```
sindresorhus: What link?

docs/rules/no-import-test-files.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Fail
+
+```js
+// File: src/index.js
+// Invalid because *.test.js is considered as a test file.
+import tests from './index.test.js';
+
+// File: src/index.js
```
sindresorhus: It's confusing that you have multiple file examples in the same code block. They should be separate code blocks.

rules/no-import-test-files.js
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	return {
+		ImportDeclaration: node => {
+			const isImportingTestFile = isTestFile(files, projectInfo.rootDir, filename, node.source.value);
+
+			if (isImportingTestFile) {
+				context.report({
+					node,
+					message: `You are importing a test file`,
```
sindresorhus: > Test files should not be imported

rules/no-import-test-files.js
```diff
@@ -0,0 +1,100 @@
...
+				});
+			}
+		},
+		CallExpression: node => {
+			if (node.callee.type !== 'Identifier' || node.callee.name !== 'require') {
```
sindresorhus: It's clearer to write it like this:


```js
if (!(node.callee.type === 'Identifier' && node.callee.name === 'require')) {
```

Same with the below.

rules/no-import-test-files.js
```diff
@@ -0,0 +1,100 @@
...
+	return {
+		ImportDeclaration: node => {
+			const isImportingTestFile = isTestFile(files, projectInfo.rootDir, filename, node.source.value);
+
+			if (isImportingTestFile) {
+				context.report({
+					node,
+					message: `You are importing a test file`,
+				});
+			}
```
sindresorhus: Line 50 to 57 is almost duplicated with the `CallExpression` one. Just make an abstracted method that accepts the `value`.

rules/no-import-test-files.js
```diff
@@ -0,0 +1,100 @@
...
+	};
+};
+
+const schema = [{
+	type: 'object',
+	properties: {
+		files: {
+			anyOf: [
+				{type: 'array'},
+				{type: 'string'}
```
sindresorhus: `string` should not be allowed

index.js
```diff
@@ -37,7 +37,8 @@ module.exports = {
 				'ava/use-t-well': 'error',
 				'ava/use-t': 'error',
 				'ava/use-test': 'error',
-				'ava/use-true-false': 'error'
+				'ava/use-true-false': 'error',
+				'ava/no-import-test-files': 'error',
```
sindresorhus: This needs to be in the same position as placed in the readme.

sindresorhus:
Nice work again, @HenriBeck 🙌

novemberborn:
Thanks @HenriBeck! If you're up for another challenge, we should support custom extensions with this rule. See https://github.com/avajs/eslint-plugin-ava/issues/201.

HenriBeck:
Will check it out next week.

PR #194, opened 2018-06-22
Fix `use-t` failing on macros

### LINKED ISSUE #127: use-t with macros
I have a macro test which takes a function argument:

``` js
test('with the error if fn throws', asapInvokesCrashHandler, err => { throw err })
```

`use-t` complains about `err` not being `t`. I think it should check whether it is the first argument, or if it's the second, whether the first argument could be a string. That should prevent false positives.

emarcotte:
As mentioned in #127, calls to test using macros would treat any function arguments to the macro as needing to meet the requirements of the use-t validation.

This change will evaluate functions for the "t" parameter only if the function is the first or second position of the call to test.

Fixes #127

### DIFF
--- a/rules/use-t.js
+++ b/rules/use-t.js
@@ -11,7 +11,12 @@ const create = context => {
 			ava.isInTestFile,
 			ava.isTestNode
 		])(node => {
-			const functionArg = node.arguments[node.arguments.length - 1];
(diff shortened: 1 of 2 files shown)


sindresorhus:
Thank you :)

PR #198, opened 2018-07-06
Disallow import {test} from ava

### LINKED ISSUE #197: use-test rule should not allow named imports
I think this shouldn't be allowed:
```js
import {test} from 'ava';
```
Import default should be forced:
```js
import test from 'ava';
```
Ava works fine if you name-import it but some other `eslint-plugin-ava` rules don't work properly if you name-import ava

atabel:
fixes #197

### DIFF
--- a/docs/rules/use-test.md
+++ b/docs/rules/use-test.md
@@ -11,6 +11,7 @@ var ava = require('ava');
 let ava = require('ava');
 const ava = require('ava');
 import ava from 'ava';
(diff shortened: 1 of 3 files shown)


atabel:
WIP, I want to make this autofixable

atabel:
auto-fix done ;)

novemberborn:
Let's see if we should remove that export from AVA instead: https://github.com/avajs/ava/issues/1862

PR #217, opened 2019-03-16
Fix async function with for-await-of loop being reported

### LINKED ISSUE #216: `no-async-fn-without-await` rule does not support async iteration (for-await-of loop)
The following is reported even though the function uses `await`:

```js
test('foobar', async t => { // Function was declared as `async` but doesn't use `await`
	for await (const bar of foo) {
		console.log(bar)
	}
});
```

futpib:
Fixes #216


novemberborn:
Nice catch, thanks @futpib!

Issue #213, opened 2019-01-13
Exception on invalid argument to `require`.


coreyfarrell:
[<img alt="Issuehunt badges" src="https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/213/badge.svg" />](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/213)

The following code causes `xo` to crash:
```js
const something = require(true);
```

Obviously this is not a valid thing to do, but it should cause eslint to report an error instead of throwing an exception.  This is something I found with a test which verifies that my babel plugin does not blow up if a non-string is passed to `require`.  For now I've had to add the test fixture files to `xo.ignores` in my package.json.

```text
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type boolean
    at assertPath (path.js:39:11)
    at Object.resolve (path.js:1085:7)
    at isTestFile (/usr/src/npm/babel-plugin-template-html-minifier/node_modules/eslint-plugin-ava/rules/no-import-test-files.js:9:36)
    at /usr/src/npm/babel-plugin-template-html-minifier/node_modules/eslint-plugin-ava/rules/no-import-test-files.js:26:31
    at CallExpression (/usr/src/npm/babel-plugin-template-html-minifier/node_modules/eslint-plugin-ava/rules/no-import-test-files.js:65:5)
    at listeners.(anonymous function).forEach.listener (/usr/src/npm/babel-plugin-template-html-minifier/node_modules/eslint/lib/util/safe-emitter.js:45:58)
    at Array.forEach (<anonymous>)
    at Object.emit (/usr/src/npm/babel-plugin-template-html-minifier/node_modules/eslint/lib/util/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/usr/src/npm/babel-plugin-template-html-minifier/node_modules/eslint/lib/util/node-event-generator.js:251:26)
    at NodeEventGenerator.applySelectors (/usr/src/npm/babel-plugin-template-html-minifier/node_modules/eslint/lib/util/node-event-generator.js:280:22)
```

> There is a $20.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/213).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/avajs/eslint-plugin-ava/) to discover more funded issues.
[...]

PR #221, opened 2019-03-24
rule `no-import-test-files`: modules that end in .js

### LINKED ISSUE #215: no-import-test-files incorrectly flags modules that end in .js
[Issuehunt badges](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/215)

Using all the default options, the following line in a test file is incorrectly flagged as violating the no-import-test-files rule:

```javascript
var highlight = require('highlight.js')
```

It appears that any module name ending in `.js` will similarly be flagged as a violation. However, the semantics for node's `require` function will _only_ look for local files if they start with a dot, so this will always be importing a module, not a test file.

> There is a $30.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/215).

Manu1400:
Fixes #215.

- [x] add new tests
- [x] use require.resolve() function

### DIFF
--- a/rules/no-import-test-files.js
+++ b/rules/no-import-test-files.js
@@ -17,6 +17,7 @@ function isTestFile(files, rootDir, sourceFile, importedFile) {
 		} catch (error) {
 			isHavingError = true;
 		}
+
 		return isHavingError;
 	};
 
--- a/rules/no-import-test-files.js
+++ b/rules/no-import-test-files.js
@@ -1,23 +1,34 @@
 'use strict';
-const path = require('path');
+const {resolve, relative, dirname} = require('path');
 const arrify = require('arrify');
-const pkgUp = require('pkg-up');
+const {sync} = require('pkg-up');
 const multimatch = require('multimatch');
-const util = require('../util');
+const {getAvaConfig, defaultFiles, getDocsUrl} = require('../util');
 
 function isTestFile(files, rootDir, sourceFile, importedFile) {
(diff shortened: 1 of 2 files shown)


rules/no-import-test-files.js
```diff
@@ -1,23 +1,35 @@
 'use strict';
-const path = require('path');
+const {resolve, relative, dirname} = require('path');
 const arrify = require('arrify');
-const pkgUp = require('pkg-up');
+const {sync} = require('pkg-up');
 const multimatch = require('multimatch');
-const util = require('../util');
+const {getAvaConfig, defaultFiles, getDocsUrl} = require('../util');
```
novemberborn: Could you revert these changes?

novemberborn:
@Manu1400 looks like @GMartigny also took at try at solving this. His PR has fewer changes and the tests pass, would you mind having a look? #226.

For your next PR, it helps if you put something like "Fixes #215" in the PR description (I've added it above). That way it's more discoverable from the issue discussion.

novemberborn:
@Manu1400 just happened to notice this PR includes a fix for #213 as well (https://github.com/avajs/eslint-plugin-ava/pull/221/commits/2d487046c3ecbc8626057f40bf06f76d55edba08). Here too @GMartigny made a separate PR: #228. I like your approach better though, could you open a PR just for that change?

sindresorhus:
Closing this in favor of #226.

PR #222, opened 2019-03-25
Declare a JSON schema in 'default' for each rule

### LINKED ISSUE #214: Use JSON schema `default` in rules
[Issuehunt badges](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/214)

Instead of manual `Object.assign`. See: https://github.com/eslint/eslint/commit/550de1e611a1e9af873bcb18d74cf2056e8d2e1b

> There is a $40.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/214).

Manu1400:
For fix #214 :
- [x] each rule have a JSON schema

### DIFF
--- a/rules/assertion-arguments.js
+++ b/rules/assertion-arguments.js
@@ -150,6 +150,7 @@ const schema = [{
 module.exports = {
 	create,
 	meta: {
+		type: 'problem',
 		docs: {
 			url: util.getDocsUrl(__filename)
 		},
--- a/rules/assertion-arguments.js
+++ b/rules/assertion-arguments.js
@@ -141,7 +141,8 @@ const schema = [{
 			enum: [
 				'always',
 				'never'
-			]
+			],
+			default: null
 		}
 	}
 }];
--- a/rules/max-asserts.js
+++ b/rules/max-asserts.js
@@ -5,9 +5,11 @@ const createAvaRule = require('../create-ava-rule');
 
 const notAssertionMethods = ['plan', 'end'];
 
+const MAX_NUMBER_ASSERTIONS_BY_DEFAULT = 5;
+
 const create = context => {
 	const ava = createAvaRule();
-	const maxAssertions = context.options[0] || 5;
+	const maxAssertions = context.options[0] || MAX_NUMBER_ASSERTIONS_BY_DEFAULT;
 	let assertionCount = 0;
 	let nodeToReport = null;
 
@@ -55,13 +57,6 @@ const create = context => {
 	});
 };
 
-const schema = [{
-	title: 'maximum number of assertions for each test',
-	type: 'integer',
-	minimum: 0,
-	default: 5
-}];
-
 module.exports = {
 	create,
 	meta: {
@@ -69,6 +64,11 @@ module.exports = {
 		docs: {
 			url: util.getDocsUrl(__filename)
 		},
-		schema
+		schema: [{
+			title: 'Maximum number of assertions for each test',
+			type: 'integer',
+			minimum: 0,
+			default: MAX_NUMBER_ASSERTIONS_BY_DEFAULT
+		}]
 	}
 };
--- a/rules/max-asserts.js
+++ b/rules/max-asserts.js
@@ -65,6 +65,7 @@ const schema = [{
 module.exports = {
 	create,
 	meta: {
+		type: 'suggestion',
 		docs: {
 			url: util.getDocsUrl(__filename)
 		},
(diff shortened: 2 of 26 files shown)


rules/max-asserts.js
```diff
@@ -56,12 +56,16 @@ const create = context => {
 };
 
 const schema = [{
-	type: 'integer'
+	title: 'maximum number of assertions for each test',
+	type: 'integer',
+	minimum: 0,
+	default: 5
```
sindresorhus: When you define the defaults here, you also need to remove the default handling from the rule itself. For example: https://github.com/avajs/eslint-plugin-ava/blob/d812210ea6d21dba0aabfbbed1e7a60dbaec5c25/rules/max-asserts.js#L10

rules/max-asserts.js
```diff
@@ -56,12 +56,16 @@ const create = context => {
 };
 
 const schema = [{
-	type: 'integer'
+	title: 'maximum number of assertions for each test',
```
sindresorhus:
```suggestion
	title: 'Maximum number of assertions for each test',
```

rules/no-async-fn-without-await.js
```diff
@@ -51,8 +51,10 @@ const create = context => {
 module.exports = {
 	create,
 	meta: {
+		type: 'problem',
 		docs: {
 			url: util.getDocsUrl(__filename)
-		}
+		},
+		schema: []
```
sindresorhus: No point in defining this property when it's just an empty array.

rules/no-import-test-files.js
```diff
@@ -69,16 +69,28 @@ const create = context => {
 };
 
 const schema = [{
+	title: 'one option: files',
```
sindresorhus: This is not a good title.

rules/no-import-test-files.js
```diff
@@ -69,16 +69,28 @@ const create = context => {
...
+	$comment: 'the default value of option `files` is from package.json or ava.config.js',
```
sindresorhus:
```suggestion
	$comment: 'The default value of the `files` option is from `package.json` or `ava.config.js`.',
```

rules/no-import-test-files.js
```diff
@@ -69,16 +69,28 @@ const create = context => {
...
+			type: 'array',
+			minItems: 0,
+			uniqueItems: true,
+			items: {
+				type: 'string'
+			},
+			examples: [
+				['lib/**/*.test.js', 'utils/**/*.test.js']
+			],
+			default: null
```
sindresorhus: `undefined` would be more correct.

rules/no-import-test-files.js
```diff
@@ -69,16 +69,28 @@ const create = context => {
...
 		}
 	}
 }];
 
 module.exports = {
 	create,
+	type: 'problem',
```
sindresorhus: ?

rules/no-identical-title.js
```diff
@@ -62,8 +62,10 @@ const create = context => {
 module.exports = {
 	create,
 	meta: {
+		type: 'suggestion',
```
sindresorhus: This is actually a `problem` as AVA will not allow tests with identical titles.

sindresorhus:
Bump :)

Manu1400:
this PR is ready

rules/max-asserts.js
```diff
@@ -5,9 +5,11 @@ const createAvaRule = require('../create-ava-rule');
 
 const notAssertionMethods = ['plan', 'end'];
 
+const MAX_NUMBER_ASSERTIONS_BY_DEFAULT = 5;
+
 const create = context => {
 	const ava = createAvaRule();
-	const maxAssertions = context.options[0] || 5;
+	const maxAssertions = context.options[0] || MAX_NUMBER_ASSERTIONS_BY_DEFAULT;
```
sindresorhus: Why are you defining the default both here and in the schema?

rules/no-ignored-test-files.js
```diff
@@ -94,6 +105,7 @@ module.exports = {
 		docs: {
 			url: util.getDocsUrl(__filename)
 		},
+		type: 'problem',
```
sindresorhus: Should be placed above `docs`. Same with the other places.

rules/assertion-arguments.js
```diff
@@ -141,14 +141,16 @@ const schema = [{
 			enum: [
 				'always',
 				'never'
-			]
+			],
+			default: null
```
sindresorhus: `undefined` would be better

PR #225, opened 2019-03-28
Make the `no-import-test` rule take into account AVA `extensions` and `babel` config

### LINKED ISSUE #201: Consider extensions in no-import-test-file rule
[Issuehunt badges](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/201)

The `no-import-test-file` rule does not consider the `extensions` and `babel.extensions` options. It only works for `.js` files.

> There is a $80.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/201).

GMartigny:
fix #201


GMartigny:
Hi,
This is still a draft of a complete solution.

There are 2 high-order decisions that need to be made:
 - [ ] What are the default extensions checked by the plugin ?
       (ex: `['js', 'jsx']`)
 - [ ] Should the `extensions` option override the extensions on the `files` options ?
       (ex: `files: ['./*.test.js'], extensions: ['txt']` => `['./*.test.txt']`

novemberborn:
> * What are the default extensions checked by the plugin ?

`js` since that's AVA's default.

> * Should the `extensions` option override the extensions on the `files` options ?

Specifying the `files` option is not sufficient. The extension needs to be listed in the `extensions` options.

GMartigny:
> Specifying the files option is not sufficient. The extension needs to be listed in the extensions options.
Should the plugin throw an error or warn when the extensions defined in `files` don't match the `extensions` option ?

rules/no-import-test-files.js
```diff
@@ -43,7 +45,13 @@ const create = context => {
 
 	const projectInfo = getProjectInfo();
 	const options = context.options[0] || {};
-	const files = arrify(options.files || projectInfo.files || util.defaultFiles);
+
+	const extensions = arrify(options.extensions || (projectInfo.babel && projectInfo.babel.extensions) || util.defaultExtensions);
+
+	const files = arrify(options.files || projectInfo.files || util.defaultFiles).filter(file => {
+		const extension = path.extname(file).substr(1);
```
sindresorhus: Just my preference:

```suggestion
		const extension = path.extname(file).slice(1);
```

sindresorhus:
> Should the plugin throw an error or warn when the extensions defined in files don't match the extensions option ?

👎 I don't think we want to get into the mess of trying to parse glob patterns.

sindresorhus:

GMartigny:
> Can you fix the merge conflict?

Done

While writing the doc for the new option, I realized I filter out glob pattern. I fix this issue and update tests.

PR #226, opened 2019-03-29
Fix `no-import-test-files` rule incorrectly reporting module names ending in `.js`

### LINKED ISSUE #215: no-import-test-files incorrectly flags modules that end in .js

GMartigny:
Fix #215

### DIFF
--- a/rules/no-import-test-files.js
+++ b/rules/no-import-test-files.js
@@ -5,6 +5,11 @@ const pkgUp = require('pkg-up');
 const multimatch = require('multimatch');
 const util = require('../util');
 
+const externalModuleRegExp = /^\w/;
+function isExternalModule(name) {
+	return externalModuleRegExp.test(name);
+}
+
 function isTestFile(files, rootDir, sourceFile, importedFile) {
 	const absoluteImportedPath = path.resolve(path.dirname(sourceFile), importedFile);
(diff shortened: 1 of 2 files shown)


GMartigny:
Hi,
The `isExternalModule` function is based on https://github.com/benmosher/eslint-plugin-import/blob/master/src/core/importType.js#L40 minus the part doing a full resolve of the import. First, it feels overkill in this context. Secondly, I'm not confident doing a resolve of a user-defined path, as it open the door for script injection.

I also add the regression test of the initial use-case.

sindresorhus:
In the future, make sure you PR title describe **what** it fixes, not how.

PR #229, opened 2019-03-29
Add auto-fix of misspelled `falsey` to `falsy`

### LINKED ISSUE #190: Auto-fix `falsey` to `falsy` in the `use-t-well` rule
Context: https://github.com/avajs/ava/issues/1807

GMartigny:
fix #190

### DIFF
--- a/docs/rules/use-t-well.md
+++ b/docs/rules/use-t-well.md
@@ -4,7 +4,7 @@ Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/re
 
 Prevent the use of unknown assertion methods and the access to members other than the assertion methods and `context`, as well as some known misuses of `t`.
 
-This rule is automatically fixable, it will replace misspelled `falsey` by `falsy`.
+This rule is partly automatically fixable. It will replace misspelled `falsey` by `falsy`.
 
 
 ## Fail
--- a/docs/rules/use-t-well.md
+++ b/docs/rules/use-t-well.md
@@ -4,6 +4,8 @@ Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/re
 
 Prevent the use of unknown assertion methods and the access to members other than the assertion methods and `context`, as well as some known misuses of `t`.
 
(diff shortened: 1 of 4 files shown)


GMartigny:
Hi,

It feels like this should be in a separate rule as this is the only error "fixable".

The error message could be better, but I've gone with simplicity.

 - [ ] Still need to update the documentation

novemberborn:
Thanks for all your PRs @GMartigny!

> It feels like this should be in a separate rule as this is the only error "fixable".

@sindresorhus?

sindresorhus:
You need to update the readme and rule docs to indicate it's fixable.

sindresorhus:
> It feels like this should be in a separate rule as this is the only error "fixable".

I think it's fine to keep it in this rule.

sindresorhus:
> It feels like this should be in a separate rule as this is the only error "fixable".

https://github.com/avajs/eslint-plugin-ava/blob/ab64fde19092318aa6b7726e81bb1808d866fdba/test/use-t-well.js#L115-L116 could potentially be fixable too.

GMartigny:
> You need to update the readme and rule docs to indicate it's fixable.

I updated the readme.md, but other fixable rules (no-only-test, no-skip-test) doc don't mention they are fixable.
What could be added to better communicate the fixable state ?

sindresorhus:
> I updated the readme.md, but other fixable rules (no-only-test, no-skip-test) doc don't mention they are fixable.

That's just an oversight. Would you be willing to fix those in a separate PR?

sindresorhus:
> What could be added to better communicate the fixable state ?

Explain what's fixable in the rule docs. Example: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/prevent-abbreviations.md#prevent-abbreviations

sindresorhus:
> Explain what's fixable in the rule docs. Example: sindresorhus/eslint-plugin-unicorn:docs/rules/prevent-abbreviations.md@master#prevent-abbreviations 

@GMartigny Can you add this to the rule documentation for this rule so this PR can be merged?

PR #231, opened 2019-04-12
Prevent incorrect detection of AVA functions when accessing through "context"

### LINKED ISSUE #223: Incorrect detection of properties of t.context as AVA functions
[Issuehunt badges](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/223)

Hi,
When a property of `t.context` match an existing AVA function (say `end`), the eslint plugin fire an error for incorrect use of this function.
Consider:
```js
test.beforeEach((t) => {
    t.context = new Vector(1, 2);
});
test("constructor", (t) => {
    t.is(t.context.start, 1);
    t.is(t.context.end, 2);		// <= `t.end()` should only be used inside of `test.cb()`  ava/no-invalid-end
});
```

I have search the source and it appear that only `end` and `skip` raised the issue.

> There is a $40.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/223).

GMartigny:
Add tests for all rules using "nameOfRootObject"
Fix #223

### DIFF
--- a/rules/no-skip-assert.js
+++ b/rules/no-skip-assert.js
@@ -11,13 +11,14 @@ const create = context => {
 			ava.isInTestFile,
 			ava.isInTestNode
 		])(node => {
-			if (node.property.name === 'skip' &&
-				util.getNameOfRootNodeObject(node) === 't' &&
-				!util.isPropertyUnderContext(node)) {
-				context.report({
-					node,
-					message: 'No assertions should be skipped.'
-				});
+			if (node.property.name === 'skip') {
+				const root = util.getRootNode(node);
+				if (root.object.name === 't' && util.assertionMethods.has(root.property.name)) {
+					context.report({
+						node,
+						message: 'No assertions should be skipped.'
+					});
+				}
 			}
 		})
 	});
--- a/rules/no-skip-assert.js
+++ b/rules/no-skip-assert.js
@@ -11,7 +11,7 @@ const create = context => {
 			ava.isInTestFile,
 			ava.isInTestNode
 		])(node => {
-			if (node.property.name === 'skip' && util.nameOfRootObject(node) === 't') {
+			if (node.property.name === 'skip' && node.object.name === 't') {
 				context.report({
 					node,
 					message: 'No assertions should be skipped.'
--- a/rules/no-skip-assert.js
+++ b/rules/no-skip-assert.js
@@ -11,7 +11,9 @@ const create = context => {
 			ava.isInTestFile,
 			ava.isInTestNode
 		])(node => {
-			if (node.property.name === 'skip' && node.object.name === 't') {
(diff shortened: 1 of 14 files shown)


GMartigny:
Hi,

I have not added tests to the `assertion-message` rule because it's deprecated and because I didn't want to add a hard-coded list of supported AVA functions.

I've also done a slight refactoring on util. If you don't want to include a refactoring to a fix, I'll gladly revert this part.

util.js
```diff
@@ -22,20 +22,20 @@ const defaultFiles = [
...
-exports.isInContext = node => {
-	if (node.object.type === 'MemberExpression') {
-		return exports.isInContext(node.object);
-	}
+exports.nameOfRootObject = node => {
+	return exports.getRoot(node).object.name;
+};
 
-	return node.property.name === 'context';
+exports.isInContext = node => {
```
sindresorhus: `isInContext` is too generic. The name needs to be more specific.

util.js
```diff
@@ -22,20 +22,20 @@ const defaultFiles = [
 	'**/*.test.js'
 ];
 
-exports.nameOfRootObject = node => {
+exports.getRoot = node => {
```
sindresorhus: `getRoot` of what? The name should be more specific.

rules/no-skip-assert.js
```diff
@@ -11,7 +11,7 @@ const create = context => {
 			ava.isInTestFile,
 			ava.isInTestNode
 		])(node => {
-			if (node.property.name === 'skip' && util.nameOfRootObject(node) === 't') {
+			if (node.property.name === 'skip' && node.object.name === 't') {
```
sindresorhus: What if it's `t.true.skip()`?

util.js
```diff
@@ -22,20 +22,20 @@ const defaultFiles = [
...
 
-	return node.object.name;
+	return node;
 };
 
-exports.isInContext = node => {
-	if (node.object.type === 'MemberExpression') {
-		return exports.isInContext(node.object);
-	}
+exports.nameOfRootObject = node => {
```
sindresorhus:
```suggestion
exports.nameOfPropertyRootObject = node => {
```

?

rules/no-skip-assert.js
GMartigny: `skip` can be before or after assertion (`t.is.skip` or `t.skip.is`) so I fixed it but `end`, `true` and `false` should be considered only when following `t` (`t.end` and not `t.whatever.end`).

novemberborn: It can only be after.

GMartigny: Ok, my bad. Unit tests are broken then.

PR #232, opened 2019-04-12
Support named imports and destructured requires

### LINKED ISSUE #121: Handle destructured import of a test modifier
[Issuehunt badges](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/121)

From https://github.com/sindresorhus/conf/issues/2#issuecomment-227897259

I think we should handle the following cases:

``` js
import {serial as test} from 'ava';
```

``` js
import {serial} from 'ava';
```

It's a useful pattern to make a test file with only serial tests.

> There is a $80.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/121).

alexzherdev:
Fixes #121

### DIFF
--- a/create-ava-rule.js
+++ b/create-ava-rule.js
@@ -193,7 +193,7 @@ module.exports = () => {
 			}
 		},
 		VariableDeclarator: node => {
-			if (!isTestFile && avaVariableDeclaratorAsts.some(ast => console.log(JSON.stringify(espurify(node), null, 2)) || deepStrictEqual(espurify(node), ast))) {
+			if (!isTestFile && avaVariableDeclaratorAsts.some(ast => deepStrictEqual(espurify(node), ast))) {
 				isTestFile = true;
 			}
 		},
--- a/create-ava-rule.js
+++ b/create-ava-rule.js
@@ -4,7 +4,7 @@ const enhance = require('enhance-visitors');
 const deepStrictEqual = require('deep-strict-equal');
 const util = require('./util');
 
-const avaImportDeclarationAst = {
+const avaImportDeclarationAsts = [{
 	type: 'ImportDeclaration',
 	specifiers: [
 		{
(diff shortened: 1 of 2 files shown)


create-ava-rule.js
```diff
@@ -19,9 +19,66 @@ const avaImportDeclarationAst = {
 		type: 'Literal',
 		value: 'ava'
 	}
-};
+}, {
+	type: 'ImportDeclaration',
```
alexzherdev: Not happy with the wall of ASTs... any way to make this easier to comprehend?

alexzherdev:
Tested this locally in a random rule test. Should we create a dedicated test for `create-ava-rule.js`?

create-ava-rule.js
```diff
@@ -65,12 +188,12 @@ module.exports = () => {
...
 	const predefinedRules = {
 		ImportDeclaration: node => {
-			if (!isTestFile && deepStrictEqual(espurify(node), avaImportDeclarationAst)) {
+			if (!isTestFile && avaImportDeclarationAsts.some(ast => deepStrictEqual(espurify(node), ast))) {
 				isTestFile = true;
 			}
 		},
 		VariableDeclarator: node => {
-			if (!isTestFile && deepStrictEqual(espurify(node), avaVariableDeclaratorAst)) {
+			if (!isTestFile && avaVariableDeclaratorAsts.some(ast => console.log(JSON.stringify(espurify(node), null, 2)) || deepStrictEqual(espurify(node), ast))) {
```
novemberborn: Did you mean to commit the `console.log()` call?

alexzherdev: Oops 🤦‍♂️ fixed

sindresorhus:
> Tested this locally in a random rule test. Should we create a dedicated test for create-ava-rule.js?

Yes, this needs to be tested.

create-ava-rule.js
sindresorhus: You might have been able to reduce the amount of AST code if you did the checks programmatically instead of comparing the AST tree, but that will also be more fragile and IMHO less readable, so I think this is fine.

sindresorhus:
@alexzherdev Would you be willing to help review some of the other PRs? 🙏

alexzherdev:
> @alexzherdev Would you be willing to help review some of the other PRs? 🙏

sure 👌

GMartigny:
@alexzherdev You still need to add test for this. Maybe you can use what's already done to test rules to be in a "real life" scenario.

sindresorhus:
@alexzherdev Still interested in finishing this? :)

alexzherdev:
☝️updated

create-ava-rule.js
GMartigny: Some comments here and there to define what each AST match against could help.

test/create-ava-rule.js
```diff
@@ -0,0 +1,65 @@
...
+
+const ruleTester = avaRuleTester(test, {
+	env: {
+		es6: true,
+	},
+	parserOptions: {
+		sourceType: 'module'
+	}
+});
+
```
sindresorhus:
```js
const errors = [
	{
		message: 'not a test file'
	}
];
```

and then reuse it.

PR #233, opened 2019-04-12
Allow the use of `t.title` with the `use-t-well` rule

### LINKED ISSUE #176: Allow use of `t.title` for `use-t-well`
I have fixtures sorted by test title, and it would be nice if i could use `t.title` in a macro to look up which fixture applies to the test, rather than repeating the title somewhere in the body of the test.

Also, if this would become a supported part of the ava api, it would be nice to have it on ava's README.md.

Please? 😍

GMartigny:
Fix #176


sindresorhus:
Thanks :)

PR #234, opened 2019-04-14
Calling a `.cb` test with the first argument being a reference disables the rule

### LINKED ISSUE #119: test-ended rule does not like macros:
``` js
function cbMacro(t, input, expected) {
  fn(input, output => {
     t.is(output, expected);
     t.end();
  });
}

test.cb(cbMacro, 'foo', 'FOO'); // <-- violates the rule, but it shouldn't.
test.cb(cbMacro, 'bar', 'BAR');
```

GMartigny:
Fix #119

### DIFF
--- a/rules/test-ended.js
+++ b/rules/test-ended.js
@@ -5,28 +5,40 @@ const util = require('../util');
 
 const create = context => {
 	const ava = createAvaRule();
+	const hasCbModifier = () => ava.hasTestModifier('cb');
 	let endCalled = false;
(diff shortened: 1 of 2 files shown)


sindresorhus:
@GMartigny You need to submit the PR URL here: https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/119

GMartigny:
Yep done, thanks for the heads up.

PR #240, opened 2019-05-03
Add `no-error-ctor-with-notthrows` rule

### LINKED ISSUE #178: Prevent specifying error type in `t.notThrows()`
[Issuehunt badges](https://oss.issuehunt.io/r/avajs/eslint-plugin-ava/issues/178)

It's an easy mistake to make if you have many `t.throws()` statements and then some `t.notThrows()` statements.

For example: https://github.com/sindresorhus/p-queue/commit/80e8e25926a6632d0429ebb526ee78a68f01d565

```js
// Fail
t.notThrows(() => {
 t.pass();
}, TypeError);

// Ok
t.notThrows(() => {
 t.pass();
});
```

It should fail if it's a variable matching this regex: `/^(?:[A-Z][a-z0-9]*)*Error$/`

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $60.00)

- [issuehunt issuehunt](https://oss.issuehunt.io/u/issuehunt) ($60.00)

### Submitted pull Requests
- [#240 rule `prevent-errortype`](https://oss.issuehunt.io/r/avajs/eslint-plugin-ava/pull/240)
- [#326 Add `no-error-ctor-with-notthrows` rule](https://oss.issuehunt.io/r/avajs/eslint-plugin-ava/pull/326)
- [#835 Add `request` and `options` to `beforeError` hook state](https://oss.issuehunt.io/r/sindresorhus/ky/pull/835)
---

### Tips

- Checkout the [Issuehunt explorer](https://oss.issuehunt.io/r/avajs/eslint-plugin-ava/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://oss.issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

gurrrung:
fixes #178 
added rule with tests (100% coverage).


sindresorhus:
Please read and adhere to https://github.com/avajs/eslint-plugin-ava/blob/master/docs/new-rule.md (Probably smart to read it multiple times as it's easy to miss something).

The rule needs a more specific name that follows the naming convention of the other rules.

rules/prevent-errortype.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+const {visitIf} = require('enhance-visitors');
+const util = require('../util');
+const createAvaRule = require('../create-ava-rule');
+
+const nameRegexp = /^(?:[A-Z][a-z0-9]*)*Error$/;
```
sindresorhus:
```suggestion
const errorNameRegex = /^(?:[A-Z][a-z\d]*)*Error$/;
```

test/prevent-errortype.js
```diff
@@ -0,0 +1,32 @@
...
+	env: {
+		es6: true
+	}
+});
+
+const errors = [{ruleId: 'prevent-errortype'}];
+
+const header = 'const test = require(\'ava\');\n';
+
+ruleTester.run('prevent-errortype', rule, {
```
sindresorhus: Needs some more tests. For example, one where the second argument is `true`, `string`, and `object`.

test/prevent-errortype.js
```diff
@@ -0,0 +1,32 @@
...
+		header + 'test(t => {t.throws(() => {t.pass();}, TypeError);});',
+		header + 'test(t => {t.end(); });'
+	],
+	invalid: [
+		{
+			code: header + 'test(t => {t.notThrows(() => {t.pass();}, TypeError);});',
+			errors
+		},
+		{
+			code: header + 'test(\'some test\',t => {t.notThrows(() => {t.pass();}, TypeError);});',
```
sindresorhus: These would be more readable if you use a template literal. You could then also use interpolation for `header`.

rules/prevent-errortype.js
```diff
@@ -0,0 +1,50 @@
...
+				return;
+			}
+
+			const calleeProperty = node.callee.property.name;
+
+			if (functionArgIndex === 1) {
+				functionArgName = node.arguments[1].name;
+			} else {
+				return;
+			}
```
sindresorhus: Would be clearer to invert the if-condition and use a early-return here.

gurungrahul2:
@novemberborn @sindresorhus can you guys suggest a better name? I think we can go with - `no-errortype-with-tnotthrows`. It's a little long name but is quite self-explanatory.

test/prevent-errortype.js
gurungrahul2: added.

test/prevent-errortype.js
gurungrahul2: done.

rules/prevent-errortype.js
gurungrahul2: done.

gurungrahul2:
@sindresorhus PTAL

gurungrahul2:
@sindresorhus

gurungrahul2:
@sindresorhus

novemberborn:
@gurungrahul2 thanks for the updates. We do get notifications when commits are pushed so there's no need to remind us. It just may take some time before we can get back to you.

gurungrahul2:
Sorry @novemberborn , i will take care.

docs/rules/no-typeerror-with-notthrows.md
```diff
@@ -0,0 +1,35 @@
+# Prevent specifying error type in `t.notThrows()`
+
+AVA will fail if error type is specified with `t.notThrows()`.
```
sindresorhus:
```suggestion
AVA will fail if error constructor is specified in the second argument of `t.notThrows()`.
```

index.js
```diff
@@ -36,6 +36,7 @@ module.exports = {
 				'ava/no-statement-after-end': 'error',
 				'ava/no-todo-implementation': 'error',
 				'ava/no-todo-test': 'warn',
+				'ava/no-typeerror-with-notthrows': 'error',
```
sindresorhus: This rule name doesn't make sense as it's not just about `TypeError`, but any error.

sindresorhus:
You missed this:

> (The description should be the same as the heading of the documentation file).
https://github.com/avajs/eslint-plugin-ava/blob/master/docs/new-rule.md

sindresorhus:
The rule should handle `.notThrowsAsync` too.

rules/no-typeerror-with-notthrows.js
```diff
@@ -0,0 +1,49 @@
...
+				return;
+			}
+
+			const functionArgName = node.arguments[1].name;
+
+			if (calleeProperty === 'notThrows') {
+				if (errorNameRegex.test(functionArgName)) {
+					context.report({
+						node,
+						message: 'Do not specify Error in t.notThrows()'
```
sindresorhus:
```suggestion
						message: 'Do not specify an error constructor in the second argument of t.notThrows()'
```

sindresorhus:
With https://github.com/avajs/eslint-plugin-ava/pull/240#discussion_r281024545, I meant they would be more readable if you format them like normal code, using multiple lines.

index.js
gurungrahul2: @sindresorhus  how about just `no-error-with-notthrows` ?

rules/no-typeerror-with-notthrows.js
gurungrahul2: done

gurrrung:
Can this be more better?

sindresorhus:
I haven't forgotten about this. I just haven't had the chance to review it properly yet. Maybe @JLHwung and @GMartigny have time to help review.

readme.md
```diff
@@ -79,6 +80,7 @@ The rules will only activate in test files.
 - [no-async-fn-without-await](docs/rules/no-async-fn-without-await.md) - Ensure that async tests use `await`.
 - [no-cb-test](docs/rules/no-cb-test.md) - Ensure no `test.cb()` is used.
 - [no-duplicate-modifiers](docs/rules/no-duplicate-modifiers.md) - Ensure tests do not have duplicate modifiers.
+- [no-error-ctor-with-notthrows](docs/rules/no-error-ctor-with-notthrows.md) - Prevent use of typerror with notthrows.
```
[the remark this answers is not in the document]
gurrrung: Thanks, for reviewing the PR, I have pushed a commit with recommended changes.

rules/no-error-ctor-with-notthrows.js
```diff
@@ -0,0 +1,50 @@
...
+
+const create = context => {
+	const ava = createAvaRule();
+
+	return ava.merge({
+		CallExpression: visitIf([
+			ava.isInTestFile,
+			ava.isInTestNode
+		])(node => {
+			const functionArgIndex = node.arguments.length - 1;
```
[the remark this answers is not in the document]
gurrrung: Solved in the latest commit.

rules/no-error-ctor-with-notthrows.js
```diff
@@ -0,0 +1,50 @@
...
+				return;
+			}
+
+			const functionArgName = node.arguments[1].name;
+
+			if (calleeProperty === 'notThrows' || calleeProperty === 'notThrowsAsync') {
+				if (errorNameRegex.test(functionArgName)) {
+					context.report({
+						node,
+						message: 'Do not specify an error constructor in the second argument of t.notThrows()'
```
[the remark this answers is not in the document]
gurrrung: Solved in latest commit.

test/no-error-ctor-with-notthrows.js
```diff
@@ -0,0 +1,122 @@
...
+				t.pass();
+			});
+		});`,
+
+		`${header}
+		test('some test',t => {
+			t.notThrowsAsync(() => {
+				t.pass();
+			}, {firstName:'some', lastName: 'object'});
+		});`
```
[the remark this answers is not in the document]
gurrrung: Solved in latest commit.

rules/no-error-ctor-with-notthrows.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+const {visitIf} = require('enhance-visitors');
+const util = require('../util');
+const createAvaRule = require('../create-ava-rule');
+
+const errorNameRegex = /^(?:[A-Z][a-z\d]*)*Error$/;
```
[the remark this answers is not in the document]
gurrrung: Solved in latest commit.

rules/no-error-ctor-with-notthrows.js
```diff
@@ -0,0 +1,50 @@
...
+				return;
+			}
+
+			const functionArgIndex = node.arguments.length - 1;
+
+			if (functionArgIndex !== 1) {
+				return;
+			}
+
+			const calleeProperty = node.callee.property.name;
```
JLHwung: Consider also check 
i) `node.callee.type` is of type `MemberExpression`, which also implies `node.callee.property` is not `undefined`
ii) `util.getNameOfRootNodeObject(node.callee) === 't';`

The following example explains why the checks above should be practitioned
```js
const { notThrows } = require("./my-custom-not-throws");
notThrows(foo);
```

rules/no-error-ctor-with-notthrows.js
```diff
@@ -0,0 +1,50 @@
...
+			const functionArgName = node.arguments[1].name;
```
JLHwung: We should do nothing when `node.arguments.length < 2`, there will be other rules take care of the `t.notThrows` signature.

Also, we should check if `node.arguments[1]` is an `Identifier`, hence we retrieve the identifier name without ambiguity.

rules/no-error-ctor-with-notthrows.js
```diff
@@ -0,0 +1,50 @@
...
+
+			if (calleeProperty === 'notThrows' || calleeProperty === 'notThrowsAsync') {
+				if (errorNameRegex.test(functionArgName)) {
```
JLHwung: I think we can simplify this to 
```js
functionArgName.endsWith("Error")
```

`errorNameRegex` implies that a valid identifier name will be of the format `^([A-Z][a-z\d]*)*`. However it is not sufficient, per [ECMAScript](https://www.ecma-international.org/ecma-262/#prod-IdentifierName), the following identifier name is also valid
```js
var $DOMError;
var \u{1F431}Error;
```
Instead of we match the identifierName via syntax, we can simply test if it ends up with `Error`.

rules/no-error-ctor-with-notthrows.js
gurrrung:
```
valid: [
    `const { notThrows } = require("./my-custom-not-throws");
     notThrows(foo);`
]
```

Hey, thanks for reviewing the PR. But this already passes the tests for me, can you explain a bit about how it will really be needed. Sorry if I miss silly things, just learning. :)

rules/no-error-ctor-with-notthrows.js
gurrrung: Isn't that case covered under
```
const functionArgIndex = node.arguments.length - 1;
// Here we are ignoring all other cases which dont have two arguments.
if (functionArgIndex !== 1) {
	return;
}
```
I'm curious, Can you give a failing case for ambiguity.

rules/no-error-ctor-with-notthrows.js
gurrrung: For cases like,
```
valid: const test = require('ava');
  
  		test('some test',t => {
  			t.notThrows(() => {
  				t.pass();
  			}, {firstName:'some', lastName: 'object'});
  		});
```
it will throw error, as `functionArgName` is `undefined` here and regex method won't fail in this case as it results into false.

How about keeping regex here and also checking if `functionArgName` is not undefined ( like you mentioned above ), would be both efficient and less error-prone?

rules/no-error-ctor-with-notthrows.js
JLHwung: > const functionArgIndex = node.arguments.length - 1;

Sorry I miss this line. It should be good to go.

> Can you give a failing case for ambiguity.

Sure, hence `node.arguments[1]` might not be an `Identifier` in some cases, e.g.
```
t.notThrows(() => {
        t.pass();
    }, void 0);
```
`node.arguments[1]` is a `UnaryExpression`, this node does not have a `.name` property. Instead of throwing error, I think we should skip it as it should be handled by other rules.

rules/no-error-ctor-with-notthrows.js
JLHwung: >  this already passes the tests for me

The test passed because `ava.isInTestFile` is false and the whole node would not be visited if one do not wrap them in an ava test node. So here are the example of the two checks I mentioned above
1) `node.callee.type` is not an `MemberExpression`
```js

const { notThrows } = require("./my-custom-not-throws")
${header}
		test('some test',t => {
			notThrows(foo);
		});`,
```
2) `util.getNameOfRootNodeObject(node.callee)` is not `t`

```js
const myCustomNotThrows = require("./my-custom-not-throws")
${header}
		test('some test',t => {
			myCustomNotThrows.notThrows(foo);
		});`,
```

Feel free to ask any questions as everyone is learning new things. It is great to explain into details when you got response.

rules/no-error-ctor-with-notthrows.js
JLHwung: I think regex would miss some use cases when user has different name style than CamelCase. Anyway I have also constructed exotic error class name that the regex would not throw, i.e. `$DOMError` and `\u{1F431}Error`.

Per [ESTree spec](https://github.com/estree/estree/blob/master/es5.md), only `Identifier` node will have `name` property, so I think we can check if `name` is defined, which is equivalent to check if `node` is Identifier.

In all, my advice on the checking process will be:

1) check AST pattern on `t.notThrows` and `t.throws` call expression. Skip on unmatched AST shapes.
2) check their signatures. Skip on invalid ones.
3) test if `functionArgName ` ends with error, which I think is simpler, or test if it complies to a [complex identifier name regex](https://github.com/babel/babel/blob/v7.5.4/packages/babel-parser/src/util/identifier.js#L71) for spec compliance. Throw if the test returns true.

rules/no-error-ctor-with-notthrows.js
gurrrung: > The test passed because `ava.isInTestFile` is false and the whole node would not be visited if one do not wrap them in an ava test node. So here are the 

This is what confuses me if ava automatically protects us then is it important to do these checks?

rules/no-error-ctor-with-notthrows.js
gurrrung: Yeah, I agree with this, adding it to the upcoming commit. Thanks.

rules/no-error-ctor-with-notthrows.js
gurrrung: Yes, let's go with it.

rules/no-error-ctor-with-notthrows.js
gurrrung: ping.

JLHwung: Thanks for the ping. I had a heavy workload recently.

> This is what confuses me if ava automatically protects us then is it important to do these checks?

`ava.isInTestFile` ([source](https://github.com/avajs/eslint-plugin-ava/blob/ff4f7606134ee167d0330284d59110033b3287b9/create-ava-rule.js#L45)) and `ava.isInTestNode` ([source](https://github.com/avajs/eslint-plugin-ava/blob/ff4f7606134ee167d0330284d59110033b3287b9/create-ava-rule.js#L80))limits our eslint rules only in the following scope.

For example, let's say you want to create an ava eslint rule `no-foo-call-in-test ` which disallow the usage of `foo()` inside the test implementation.
```js
test('some test',t => {  // <-- Am I inside an ava test? Looks good, let's check.
	foo() // <-- bail!!
});
foo()
// <-- I will not check the code here since I am not in ava test node.
```

So in your case, we are checking if a `CallExpression` is `t.notThrows()`, we can not simply assume that every `CallExpression` with the name `notThrows` is a valid ava call. In the true negative example I mentioned above,
```js
const { notThrows } = require("./my-custom-not-throws")
test('some test',t => { // <-- Am I inside an ava test? Looks good, let's check.
	notThrows(foo);  // <-- Is it a CallExpression? Yes. Is identifier name `notThrows`? Yes, we bail here
});

notThrows(foo); // <-- I will not check this node since I am not in a ava test.
```
Apparently it is not `t.notThrows` call so we should mark it as valid.

gurrrung: Thanks. It is done, can you double-check? @JLHwung

gurrrung:
Review ping. :) @sindresorhus

readme.md
```diff
@@ -84,6 +85,7 @@ The rules will only activate in test files.
 - [no-async-fn-without-await](docs/rules/no-async-fn-without-await.md) - Ensure that async tests use `await`.
 - [no-cb-test](docs/rules/no-cb-test.md) - Ensure no `test.cb()` is used.
 - [no-duplicate-modifiers](docs/rules/no-duplicate-modifiers.md) - Ensure tests do not have duplicate modifiers.
+- [no-error-ctor-with-notthrows](docs/rules/no-error-ctor-with-notthrows.md) - No specifying error type in `t.notThrows()`.
```
GMartigny, line 88:
```suggestion
- [no-error-ctor-with-notthrows](docs/rules/no-error-ctor-with-notthrows.md) - Ensure no error constructor is specified in `t.notThrows()`.
```

test/no-error-ctor-with-notthrows.js
```diff
@@ -0,0 +1,210 @@
...
+		es6: true
+	},
+	parserOptions: {
+		ecmaVersion: 2019
+	}
+});
+
+const errors = [{ruleId: 'no-error-ctor-with-notthrows'}];
+
+const header = `const test = require('ava');\n`; // eslint-disable-line quotes
```
GMartigny, line 16:
```suggestion
const header = 'const test = require(\'ava\');\n';
```

test/no-error-ctor-with-notthrows.js
```diff
@@ -0,0 +1,210 @@
...
+
+ruleTester.run('no-error-ctor-with-notthrows', rule, {
+	valid: [
+		`${header}
+		test('some test',t => {
```
GMartigny, line 21:
```suggestion
		test('some test', t => {
```
Could you fix this for all the other times ?

package.json
```diff
@@ -62,7 +62,7 @@
 	"ava": {
 		"files": [
 			"!rules",
-			"test/*.js"
+			"test/*js"
```
GMartigny, line 65:
```suggestion
			"test/*.js"
```
Why change this ?

GMartigny (changes requested):

sindresorhus:
Closing for lack of activity.

PR #245, opened 2019-05-15
Add support for the new eslint `meta.type` property

### LINKED ISSUE #211: Support the new rule `meta.type` property
[Issuehunt badges](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/211)

https://eslint.org/blog/2018/11/eslint-v5.9.0-released#the-fix-type-option
https://eslint.org/docs/developer-guide/working-with-rules#rule-basics

> There is a $20.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/211).

GMartigny:
Fix #211 

Since https://github.com/avajs/eslint-plugin-ava/pull/222 looks stale (Closes #222)

Few interrogations:
 * `no-import-test-file`: this is not a problem per se, but I don't know to which point it mess with AVA
 * `no-statement-after-end`: it seems to works with statement after `t.end`, but I might be missing something as I don't know what AVA does.
 * `assertion-arguments`: too few arguments won't prevent AVA from working (no crash), but it will always fail. Is this more problem or suggestion then ?


sindresorhus:
> no-import-test-file: this is not a problem per se, but I don't know to which point it mess with AVA

`suggestion`

> no-statement-after-end: it seems to works with statement after t.end, but I might be missing something as I don't know what AVA does.

`problem` (it could cause weird issues)

> assertion-arguments: too few arguments won't prevent AVA from working (no crash), but it will always fail. Is this more problem or suggestion then ?

`problem`

sindresorhus:

PR #247, opened 2019-05-15
Add `prefer-t-regex` rule

### LINKED ISSUE #163: Rule proposal: `prefer-t-regex`
[Issuehunt badges](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/163)

Users should prefer the `regex` assertion as it's more readable and gives better output.

## Fail

```js
t.true(/\d+/.test(foo()));
```

## Pass

```js
t.regex(foo(), /\d+/);
```

Same for the negated ones, with `t.notRegex`.

---

Would also be useful to be able to detect the regex from a variable somewhere else in the scope as I often share a regex between multiple asserts:

```js
const re = /\d+/;

…

t.true(re.test(foo()));
```

---

This could be auto-fixable, right?

---

Should we also detect the following?

```js
t.true(foo().search(/\d+/));
t.truthy(foo().match(/\d+/));
```

> There is a $60.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/163).

GMartigny:
Fix #163 

Open the PR for review.
I've not done the "fixable" part yet. I feel like it's going break in a lot of corner cases.


index.js
```diff
@@ -39,6 +39,7 @@ module.exports = {
 				'ava/no-unknown-modifiers': 'error',
 				'ava/prefer-async-await': 'error',
 				'ava/prefer-power-assert': 'off',
+				'ava/prefer-t-regexp': 'error',
```
sindresorhus: The assertion is called `t.regex`, not `t.regexp`.

docs/rules/prefer-t-regex.md
```diff
@@ -0,0 +1,35 @@
+# Prefer using `t.regex()` to test regular expression
```
sindresorhus:
```suggestion
# Prefer using `t.regex()` to test regular expressions
```

sindresorhus:
I guess this is still waiting for the `I've not done the "fixable" part yet. I feel like it's going break in a lot of corner cases.` part.

rules/prefer-t-regex.js
```diff
@@ -0,0 +1,88 @@
...
+
+	return ava.merge({
+		CallExpression: visitIf([
+			ava.isInTestFile,
+			ava.isInTestNode
+		])(node => {
+			// Call a boolean assertion (eg: `.true`, `.false` ...)
+			if (node.callee.type === 'MemberExpression' &&
+				booleanTests.includes(node.callee.property.name) &&
+				util.getNameOfRootNodeObject(node.callee) === 't') {
```
sindresorhus: The above expression could be moved into a descriptive variable, like `isBooleanAssertion`. Then you can use early-return instead to reduce the nesting.

rules/prefer-t-regex.js
```diff
@@ -0,0 +1,88 @@
...
+				const arg = node.arguments[0];
+
+				// Call the `test` function
+				if (arg.type === 'CallExpression') {
```
sindresorhus: Early-return to reduce nesting.

rules/prefer-t-regex.js
```diff
@@ -0,0 +1,88 @@
...
+
+					let isRegExp = lookup.regex;
+
+					// It's not a regexp but an identifier
+					if (!isRegExp && lookup.type === 'Identifier') {
+						const reference = findReference(lookup.name);
+						isRegExp = reference.init.regex;
+					}
+
+					if (isRegExp) {
```
sindresorhus: Early-return to reduce nesting.

rules/prefer-t-regex.js
```diff
@@ -0,0 +1,88 @@
...
+							const source = context.getSourceCode();
+							return [
+								fixer.replaceText(node.callee.property, assert),
+								fixer.replaceText(arg, `${source.getText(variable)}, ${source.getText(lookup)}`)
+							];
+						};
+
+						context.report({
+							node,
+							message: 'Prefer using the `.regex` assert function.',
```
sindresorhus:
```suggestion
							message: 'Prefer using the `t.regex()` assertion.',
```

And it should show `t.notRegex()` when appropriate.

test/prefer-t-regex.js
```diff
@@ -0,0 +1,63 @@
...
+			output: header + 'test(t => t.notRegex("foo", /\\d+/));',
+			errors
+		},
+		{
+			code: header + 'test(t => t.true(/\\d+/.test(foo())));',
+			output: header + 'test(t => t.regex(foo(), /\\d+/));',
+			errors
+		},
+		{
+			code: header + 'const reg = /\\d+/;\ntest(t => t.true(reg.test(foo.bar())));',
```
sindresorhus: I think we should only make it report when the regex is a literal. Otherwise `.test()` and `.match()` are way to common method names that could be anything else.

GMartigny: For what it worth, the rule check if the reference is in scope and if it's a regexp.
For example, this is valid: `const a = "not a regexp";
test(t => t.true(a.test("foo")));`

sindresorhus: Ah, ok. Then it's ok.

sindresorhus:
Looks great! Thanks for your continued contributions 🙌

PR #248, opened 2019-05-23
Refactor test cases for readability

GMartigny:
Syntaxe in test case is a mix-bag, I guess we could gain in readability and maintainability with a big refactoring.
I'm willing to do it, but it's not a small change. I prefer to validate how we should do it before diving in.
What do you think ?


test/helpers/test-case.js
```diff
@@ -0,0 +1,27 @@
+const header = 'const test = require(\'ava\');';
+
+function testCaseBuilder(ruleId, options = []) {
+	return function testCase(content, errorMessage = [], withoutHeader = false) {
```
sindresorhus: `withoutHeader = false` is not very readable at the callsite.

https://ariya.io/2011/08/hall-of-api-shame-boolean-trap

test/helpers/test-case.js
```diff
@@ -0,0 +1,27 @@
...
+		const testFn = `
+			test(t => {
```
sindresorhus: The test should be named. AVA requires this.
```suggestion
			test('x', t => {
```

test/helpers/test-case.js
```diff
@@ -0,0 +1,27 @@
+const header = 'const test = require(\'ava\');';
+
+function testCaseBuilder(ruleId, options = []) {
```
sindresorhus: Needs a doc comment on what it does and how to use it.

test/max-asserts.js
```diff
@@ -1,80 +1,62 @@
...
 	env: {
 		es6: true
 	}
 });
 
-const errors = [{ruleId: 'max-asserts'}];
-const header = 'const test = require(\'ava\');\n';
+const ruleId = 'max-asserts';
+const basicTestCase = testCaseBuilder(ruleId);
+const lowerTestCase = testCaseBuilder(ruleId, [2]);
```
sindresorhus: Reading the code here I have no what the `[2]` means. I think the API could make it more readable.

test/max-asserts.js
```diff
@@ -1,80 +1,62 @@
...
 const ruleTester = avaRuleTester(test, {
 	env: {
 		es6: true
 	}
 });
 
-const errors = [{ruleId: 'max-asserts'}];
-const header = 'const test = require(\'ava\');\n';
+const ruleId = 'max-asserts';
+const basicTestCase = testCaseBuilder(ruleId);
```
sindresorhus: I think the basic test case can just be named `testCase`.

test/helpers/test-case.js
GMartigny: I tend to use [jsDoc](https://jsdoc.app/about-getting-started.html) for this. Do you prefer something else ?

test/helpers/test-case.js
GMartigny: Are you talking about the unnamed boolean on call-site, the double negative induce by `withoutHeader = false` or both ?
I can make the use of an object literal to avoid the unnamed option. I personally prefer optional boolean value to be `false`, but I'm ok with the opposite.

sindresorhus: Both

test/helpers/test-case.js
sindresorhus: jsDoc is fine, but I think the types are. Really just needs a normal comment explaining it.

GMartigny:
If that looks better to you, I can start to use this API on all test files.

sindresorhus:
Yeah, I like it. Go ahead.

sindresorhus:
@GMartigny Still interested in finishing this?

sindresorhus:
Closing for now, but happy to reopen if you ever have to finish this, @GMartigny  :)

PR #250, opened 2019-05-25
Rely on AVA to provide a helper for our rules

novemberborn:
This impacts no-ignored-test-files and no-import-test-files. These rules
now assume an AVA version is installed that provides a helper. This
helper can classify files according to AVA's configuration.

This removes the need to specify options for these rules.

Note that the tests use a stubbed helper.

---

AVA PR: https://github.com/avajs/ava/pull/2139


novemberborn:
Here I linked both PR's from one new project:

```
❯ npx eslint .

/private/var/folders/2_/qczp184x76b2nl034sq5hvxw0000gn/T/tmp.mDR4Y3hUm0/not-test.js
  1:1  error  Test files should not be imported  ava/no-import-test-files

/private/var/folders/2_/qczp184x76b2nl034sq5hvxw0000gn/T/tmp.mDR4Y3hUm0/test/_helper.js
  1:1  error  AVA treats this as a helper file  ava/no-ignored-test-files

✖ 2 problems (2 errors, 0 warnings)
```

novemberborn:
For the `no-import-test-files` rule, I think I need to add a `classifyImport(importPath)` method to he helper. If the import has a configured extension, it can see how it's classified, and otherwise we can check it with each of the configured extensions.

novemberborn:
> For the `no-import-test-files` rule, I think I need to add a `classifyImport(importPath)` method to he helper. If the import has a configured extension, it can see how it's classified, and otherwise we can check it with each of the configured extensions.

This has been addressed.

sindresorhus:
Can you make it only load AVA if it's in a test file? I bundle this plugin in XO which is used in non-AVA projects too.

Otherwise, this looks good.

novemberborn:
> Can you make it only load AVA if it's in a test file? I bundle this plugin in XO which is used in non-AVA projects too.

For `no-ignored-test-files` I could load lazily once in `Program:exit`, yes. But correct me if I'm wrong there's no way to do this for `no-import-test-files`. I suppose we could do it lazily once there's a non-external require?

novemberborn:
@sindresorhus is this what you had in mind?

sindresorhus:
Yes, thanks. That's exactly what I had in mind.

PR #255, opened 2019-06-02
Ensure @scoped-packages are ignored in no-import-test-files

### LINKED ISSUE #253: ava/no-import-test-files triggers for scoped packages
Using the recommended config of the plugin, `import Octokit from '@octokit/rest';` inside a file called `_stubs.js`, so an ava helper file, triggers the rule `ava/no-import-test-files` with the message "Test files should not be imported" for the whole line.

Example: https://travis-ci.org/mozillach/gh-projects-content-queue/jobs/540192593#L477

novemberborn:
Only attempt to classify imports that are file paths. Fixes #253.

### DIFF
--- a/rules/no-import-test-files.js
+++ b/rules/no-import-test-files.js
@@ -2,10 +2,8 @@
 const path = require('path');
 const util = require('../util');
 
-const externalModuleRegExp = /^\w/;
-function isExternalModule(name) {
-	return externalModuleRegExp.test(name);
-}
+// Assume absolute paths can be classified by AVA.
+const isFileImport = name => path.isAbsolute(name) || name.startsWith('./') || name.startsWith('../');
 
 const create = context => {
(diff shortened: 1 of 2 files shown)


aureooms:
Is [this issue I am having](https://travis-ci.org/aureooms/js-tape/builds/542143187#L498) related to this patch?

GMartigny:
@aureooms Your issue is due to an outdated dependency of `eslint-plugin-ava` in `xo`. Once `eslint-plugin-ava@7` is ironed out, `xo` dependencies should be updated.
You could raise and issue in the `xo` repo to track it, but it'll be done anyway IMO.

In the mean time, you can run `npm install -D eslint-plugin-ava` manually to update `eslint-plugin-ava` in your repo.

aureooms:
@GMartigny Thanks for the analysis! I'll just wait.

PR #258, opened 2019-06-05
Add defaults to schema

### LINKED ISSUE #214: Use JSON schema `default` in rules

MrHen:
Fixes #214.

Honestly, I'm not sure this issue makes a lot of sense. There were only three rules I saw that had options at all:

* `assertion-arguments`: Default is `undefined`. I added it explicitly but it seems kind of redundant.
* `max-asserts`: Cannot use a default as per https://github.com/avajs/eslint-plugin-ava/issues/214#issuecomment-478915717. I'm not a big fan of non-object options anyway. I recommend converting from the numeric option to an object which would then pick up the default. This would be a breaking change, though. Alternatively, set the default in `schema` but also use the fallback in `create`.
* `test-title-format`: Default is also `undefined`.

Am I missing something about this request? There was no `Object.assign` usage to replace.

### DIFF
--- a/rules/max-asserts.js
+++ b/rules/max-asserts.js
@@ -9,8 +9,7 @@ const notAssertionMethods = ['plan', 'end'];
 
 const create = context => {
 	const ava = createAvaRule();
-	// TODO: Remove default value once defaults can be applied to integers
-	// https://github.com/eslint/eslint/issues/11749
+	// TODO: Convert to options to an object so we can set default using JSON Schema (without using this fallback)
 	const maxAssertions = context.options[0] || MAX_ASSERTIONS_DEFAULT;
 	let assertionCount = 0;
 	let nodeToReport;
--- a/rules/max-asserts.js
+++ b/rules/max-asserts.js
@@ -9,7 +9,8 @@ const notAssertionMethods = ['plan', 'end'];
 
 const create = context => {
 	const ava = createAvaRule();
-	// TODO: Convert to options to an object so we can set default using JSON Schema (without using this fallback)
+	// TODO: Convert options to object JSON Schema default works properly
+	// https://github.com/avajs/eslint-plugin-ava/issues/260
 	const maxAssertions = context.options[0] || MAX_ASSERTIONS_DEFAULT;
 	let assertionCount = 0;
 	let nodeToReport;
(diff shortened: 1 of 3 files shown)


GMartigny:
> * `max-asserts`: Cannot use a default as per [#214 (comment)](https://github.com/avajs/eslint-plugin-ava/issues/214#issuecomment-478915717).

I opened [an issue to `eslint`](https://github.com/eslint/eslint/issues/11749) reporting this. It's working as designed for now. Either we complain to `ajv` to add this behavior or we "convert from the numeric option to an object which would then pick up the default" as you said.

sindresorhus:
> Alternatively, set the default in schema but also use the fallback in create.

👍

rules/max-asserts.js
```diff
@@ -3,11 +3,15 @@ const {visitIf} = require('enhance-visitors');
...
 const createAvaRule = require('../create-ava-rule');
 
+const MAX_ASSERTIONS_DEFAULT = 5;
+
 const notAssertionMethods = ['plan', 'end'];
 
 const create = context => {
 	const ava = createAvaRule();
-	const maxAssertions = context.options[0] || 5;
+	// TODO: Remove default value once defaults can be applied to integers
```
sindresorhus: I meant a TODO about converting to a object option and then removing the fallback.

MrHen:
@sindresorhus I think I got it right this time? Although now I'm thinking I should just file an Issue for it and include a link in the TODO comment. 😛

sindresorhus:
> Although now I'm thinking I should just file an Issue for it and include a link in the TODO comment. 

👍

sindresorhus:
@MrHen Can you submit the PR URL here: https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/214 ?

MrHen:
@sindresorhus Done. Sorry, I keep forgetting it isn't automatic.

PR #262, opened 2019-06-13
Add `no-inline-assertions` rule

### LINKED ISSUE #152: `no-inline-assertions` rule
[Issuehunt badges](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/152)

For readability reasons, I try to not use inline arrow functions when the function doesn't return, to make it clear it doesn't return. I would like to prevent assertions from being called in an inline arrow function test. This might seem like an edge-case, but I've seen it many times and have had to comment about it on multiple pull requests.
## Fail

``` js
test('foo', t => t.true(fn()));
```
## Pass

``` js
test('foo', t => {
    t.true(fn());
});
```

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $60.00)

### Tips

</details>

JLHwung:
Update (2019-06-13): I have pushed `no-inline-assertions` rule implementation, see the docs and test cases. This rule will fail

Any `MemberExpression` as the arrow function body
```js
test('foo', t => t.true(fn()));
test('foo', t => t.skip());
test('foo', t => 
    t.true(fn())
);
// ...
```

And any `BlockStatement` within the same line
```js
test('foo', t => { t.true(fn()) });
test('foo', t => { t.skip() });
```

---
(2019-06-12)
This is an early-stage design discussion, as I have the following question regards to this rule. I would also put up my thoughts and any community input is welcome.

## Background: Recap of #152
The `no-inline-assertions` was originally proposed at #152, where Sindre listed two example

### Pass
```js
test('foo', t => t.true(fn()));
```

### Failure
```js
test('foo', t => {
    t.true(fn());
});
```

## Question on extra examples
While the example above is unambiguous, there is more question to be addressed after I ponder for a while.

1. Should inline arrow function with block statement be failed?
```js
test('foo', t => { t.true(fn()) });
```

2. Should inline arrow function with a return statement be failed?
```js
test('foo', t => { return t.true(fn()) });
```

3. Should inline arrow function with assertion call as last expression be failed?

```js
test('foo', t => { t.log('Meow');t.true(fn()) });
```

Here are my thoughts on these cases,
On the first, I think it should be failed as an inline arrow function with block statement is still *inline*.
On the second one, I think there should be a separate rule, e.g. `no-return-assertions` to take care of this. Otherwise we will have also to fail the following case
```js
test('foo', t => {
  return t.true(fn());
});
```
which is, of course, not an *inline* arrow function. So rule `no-inline-assertions` should not be triggered by case 2.

On the third one, I think it should be failed as it is still an *inline* arrow function with assertion calls inside. But I am not an ava user so I can't triage the impact of the third case. Is it common? Or we can neglect this case?

The docs here reflects my thoughts and is suggested to any changes.

Fixes #152

### DIFF
--- a/docs/rules/no-inline-assertions.md
+++ b/docs/rules/no-inline-assertions.md
@@ -2,6 +2,8 @@
 
 The test implementation should not purely consist of an inline assertion as assertions do not return a value and having them inline also makes the tests less readable.
 
+This rule is fixable. It will wrap assertions into brackets `{}`. The fixer would not insert extra space or linebreaks around brackets.  
+
 
 ## Fail
 
--- a/docs/rules/no-inline-assertions.md
+++ b/docs/rules/no-inline-assertions.md
@@ -1,4 +1,4 @@
-# Ensure assertions are not called from an inline function
+# Ensure no assertions are called from inline functions
 
 Prevent assertions being called from an inline function, to make it clear that it does not return.
 
@@ -18,7 +18,7 @@ test('foo', t => { t.true(fn()) });
 ```js
 import test from 'ava';
 
-test('foo', t => 
+test('foo', t =>
 	t.true(fn())
 );
 ```
--- a/docs/rules/no-inline-assertions.md
+++ b/docs/rules/no-inline-assertions.md
@@ -1,8 +1,8 @@
-# Ensure assertions are not called from an inline arrow function
+# Ensure assertions are not called from an inline function
 
 Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/related/eslint-plugin-ava/docs/rules/no-inline-assertions.md)
 
-Prevent assertions being called from an inline arrow function, to make it clear that it does not return.
+Prevent assertions being called from an inline function, to make it clear that it does not return.
 
 ## Fail
 ```js
@@ -20,7 +20,9 @@ test('foo', t => { t.true(fn()) });
 ```js
 import test from 'ava';
 
-test('foo', t => { t.log('Meow');t.true(fn()) });
+test('foo', t => 
+	t.true(fn())
+);
 ```
 
 ## Pass
--- a/docs/rules/no-inline-assertions.md
+++ b/docs/rules/no-inline-assertions.md
@@ -2,7 +2,7 @@
 
 The test implementation should not purely consist of an inline assertion as assertions do not return a value and having them inline also makes the tests less readable.
 
-This rule is fixable. It will wrap assertions into brackets `{}`. The fixer would not insert extra space or linebreaks around brackets.  
+This rule is fixable. It will wrap the assertion in braces `{}`. It will not do any whitespace or style changes.
 
 
 ## Fail
@@ -20,6 +20,6 @@ test('foo', t => t.true(fn()));
 import test from 'ava';
 
 test('foo', t => {
-	t.true(fn())
+	t.true(fn());
 });
 ```
--- a/docs/rules/no-inline-assertions.md
+++ b/docs/rules/no-inline-assertions.md
@@ -1,4 +1,4 @@
-# Ensure no assertions are called from inline functions
+# Ensure assertions are not called from inline arrow functions
 
 Prevent assertions being called from an inline function, to make it clear that it does not return.
 
--- a/docs/rules/no-inline-assertions.md
+++ b/docs/rules/no-inline-assertions.md
@@ -1,7 +1,6 @@
 # Ensure assertions are not called from inline arrow functions
 
-Prevent assertions being called from an inline function, to make it clear that it does not return.
-
+An inline arrow function is an arrow function with concise body, i.e. `x => x` will return `x` to the caller. The test implementation, composed of a series of assertions, should not be an inline arrow function as assertions do not return.
 ## Fail
 ```js
 import test from 'ava';
--- a/docs/rules/no-inline-assertions.md
+++ b/docs/rules/no-inline-assertions.md
@@ -1,0 +1,33 @@
+# Ensure assertions are not called from an inline arrow function
+
+Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/related/eslint-plugin-ava/docs/rules/no-inline-assertions.md)
+
+Prevent assertions being called from an inline arrow function, to make it clear that it does not return.
+
+## Fail
+```js
+import test from 'ava';
+
+test('foo', t => t.true(fn()));
+```
+
+```js
+import test from 'ava';
+
+test('foo', t => { t.true(fn()) });
+```
+
+```js
+import test from 'ava';
+
+test('foo', t => { t.log('Meow');t.true(fn()) });
+```
+
+## Pass
+```js
+import test from 'ava';
+
+test('foo', t => {
+	t.true(fn())
+});
+```
--- a/docs/rules/no-inline-assertions.md
+++ b/docs/rules/no-inline-assertions.md
@@ -1,14 +1,19 @@
 # Ensure assertions are not called from inline arrow functions
 
-An inline arrow function is an arrow function with concise body, i.e. `x => x` will return `x` to the caller. The test implementation, composed of a series of assertions, should not be an inline arrow function as assertions do not return.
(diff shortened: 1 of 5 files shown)


GMartigny:
I think the 2nd example fall under the [arrow-body-style](https://eslint.org/docs/rules/arrow-body-style.html) rule of eslint.

IMO, this `no-inline-assertion` should fail no matter what, when `test` second param is in the same line.

JLHwung:
> IMO, this no-inline-assertion should fail no matter what, when test second param is in the same line.

Agree. I was worrying about `test("somethigs", (t) => { t.skip() });` before but since there is an `ava/use-t-well` rule, it should be straightforward to fail when the block statement is a one-liner.

GMartigny:
> When there is an empty test body, we should not fail

I agree, no body is not an inline body.

You forgot to add the rule to recommended config in [index.js](https://github.com/avajs/eslint-plugin-ava/blob/master/index.js).

docs/rules/no-inline-assertions.md
```diff
@@ -0,0 +1,35 @@
+# Ensure assertions are not called from an inline function
+
+Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/related/eslint-plugin-ava/docs/rules/no-inline-assertions.md)
```
JLHwung: BTW this link is a dummy french translation link. As I don't speak French (I wish I could), I hope I can connect to the people and translate this document after this PR gets merged.

GMartigny: FYI @forresst

forresst: @JLHwung Do not add this line yet. I will do the translation when this PR is merged. And I will add this link when the translation is done.

JLHwung: Removed and thanks.

readme.md
```diff
@@ -83,6 +84,7 @@ The rules will only activate in test files.
 - [no-identical-title](docs/rules/no-identical-title.md) - Ensure no tests have the same title.
 - [no-ignored-test-files](docs/rules/no-ignored-test-files.md) - Ensure no tests are written in ignored files.
 - [no-import-test-files](docs/rules/no-import-test-files.md) - Ensure no test files are imported anywhere.
+- [no-inline-assertions](docs/rules/no-inline-assertions.md) - Ensure no assertions are called from inline functions.
```
sindresorhus: This is not the exact same as the doc title

JLHwung: True, personally I prefer this one. I would change the doc title to align with it.

sindresorhus:
> Should inline arrow function with block statement be failed?

That's not an inline arrow function. And no.

> Should inline arrow function with a return statement be failed?

That's not an inline arrow function. And no, I think this would be a better fit for the `use-t-well` rule. It could validate that no assertion is returned, unless it's an assertion that returns a promise, in which, it's allowed.

> Should inline arrow function with assertion call as last expression be failed?

That's not an inline arrow function. And no.

sindresorhus:
> IMO, this no-inline-assertion should fail no matter what, when test second param is in the same line.

I disagree. That's better handled by normal ESLint style rules. The point here is to handle an *inline arrow function*, meaning `x => x` (not `x => { return x }`).

rules/no-inline-assertions.js
```diff
@@ -0,0 +1,61 @@
+'use strict';
+const {visitIf} = require('enhance-visitors');
+const createAvaRule = require('../create-ava-rule');
+const util = require('../util');
+
+function report({node, context}) {
+	context.report({
+		node,
+		message: 'Assertions should not be called from an inline function.'
```
sindresorhus: This would be more correct:

```suggestion
		message: 'The test implementation should not be just an assertion in an inline arrow function.'
```

But would be nice to find a way to shorten the above slightly.

JLHwung:
> The point here is to handle an inline arrow function, meaning x => x (not x => { return x }).

Per https://tc39.es/ecma262/#prod-ConciseBody, we should alias our definition of "inline arrow function" to "arrow function with concise body", and then we can clearly state that `x => { return x }` is _not_ a `inline arrow function`.

And the error message could be
```
The test implementation should not be a concise-body arrow function.
```

If we agree that `BlockStatement` should not be considered `inline`, we should leave it as-is and only take care of `CallExpression`.

sindresorhus:
> The test implementation should not be an inline arrow function.

👍 

> If we agree that BlockStatement should not be considered inline, we should leave it as-is and only take care of CallExpression.

👍

JLHwung:
@sindresorhus Just pushed a new version and revised test cases.

docs/rules/no-inline-assertions.md
```diff
@@ -0,0 +1,19 @@
+# Ensure no assertions are called from inline functions
```
sindresorhus:
```suggestion
# Ensure assertions are not called from inline arrow functions
```

docs/rules/no-inline-assertions.md
```diff
@@ -0,0 +1,19 @@
...
+
+Prevent assertions being called from an inline function, to make it clear that it does not return.
```
sindresorhus: This is repeating the title too much. It should also use the wording "test implementation".

test/no-inline-assertions.js
```diff
@@ -0,0 +1,44 @@
...
+		es6: true
+	}
+});
+
+const errors = [{ruleId: 'no-inline-assertions'}];
+const header = 'const test = require(\'ava\');\n';
+
+ruleTester.run('no-todo-test', rule, {
+	valid: [
+		// Shouldn't be treiggered as the test implementation is not an inline arrow function
```
sindresorhus: Typo

sindresorhus:
Can you make the rule auto-fixable?

JLHwung:
> Can you make the rule auto-fixable?

Good idea, it could be autofixable and I would look into it later. Anyway I have fixed the writing issues.

JLHwung:
@sindresorhus I have pushed a simple fixer. The fixer wraps `callExpression` inside brackets. i.e. `() => x()` will be fixed to `() => {x()}`.

I tend to keep the fixer as minimal as possible, which means the other ESLint style rules, i.e. `prettier` should take care of the formatting of the new BlockStatement, such as the extra space before or after brackets or potential line breaks around the brackets.

test/no-inline-assertions.js
JLHwung: Fixed in b07ccb5.

sindresorhus:
You need to mark it as being fixable in the readme and the rule docs.

sindresorhus:
>  I have pushed a simple fixer. The fixer wraps callExpression inside brackets. i.e. () => x() will be fixed to () => {x()}.

👍 Yup. That's the correct way to do it.

readme.md
```diff
@@ -83,6 +84,7 @@ The rules will only activate in test files.
 - [no-identical-title](docs/rules/no-identical-title.md) - Ensure no tests have the same title.
 - [no-ignored-test-files](docs/rules/no-ignored-test-files.md) - Ensure no tests are written in ignored files.
 - [no-import-test-files](docs/rules/no-import-test-files.md) - Ensure no test files are imported anywhere.
+- [no-inline-assertions](docs/rules/no-inline-assertions.md) - Ensure assertions are not called from inline functions.
```
sindresorhus: The description should be in sync with the rule docs heading.

docs/rules/no-inline-assertions.md
```diff
@@ -1,14 +1,19 @@
 # Ensure assertions are not called from inline arrow functions
 
-An inline arrow function is an arrow function with concise body, i.e. `x => x` will return `x` to the caller. The test implementation, composed of a series of assertions, should not be an inline arrow function as assertions do not return.
+The test implementation should not purely consist of an inline assertion as assertions do not return a value and having them inline also makes the tests less readable.
```
JLHwung: Way better than my writing. 😂

readme.md
JLHwung: Addressed in d2f49a4.

sindresorhus:
> You need to mark it as being fixable in **the readme** and the rule docs.

Not resolved yet.

sindresorhus:
This looks good now. Thanks for working on this rule :)

PR #264, opened 2019-06-15
Add `no-incorrect-deep-equal` rule

### LINKED ISSUE #158: Rule proposal: No deepEqual for primitives values
[Issuehunt badges](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/158)

I see a lot of of uses of `t.deepEqual` that compares to a promitive value when a simple `t.is(a, b)` or `t.true(a === b)` would work.

```js
t.deepEqual(value, 'foo');
```

I'm not sure if there is a big perf hit or something, but I think it's better to use a simpler form of assertion whenever possible. I personally like to use `t.is()` or `t.true(===)` instead (frankly, most of my assertions now are `t.true()`)

# Invalid
```js
t.deepEqual(expression, 'foo');
t.deepEqual(expression, 1);
t.deepEqual(expression, `foo${bar}`);
t.deepEqual(expression, null);
t.deepEqual(expression, undefined);
t.notDeepEqual(expression, undefined);
t.deepEqual(1, expression); // ?
```

# Valid
```js
t.deepEqual(expression, otherExpression);
t.deepEqual(expression, {});
t.deepEqual(expression, []);
t.notDeepEqual(expression, []);
```

As for the rule name, I think this only targets deepEqual and notDeepEqual, so I propose `no-simple-deep-equal` or something equivalent (but would love a better name for it).

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $60.00)

### Tips

</details>

MrHen:
Closes #158.

Does not cover `t.deepEqual(1, expression); // ?` situations. Does not fix back to `deepEqual` if it detects a non-literal in `is` -- I felt that would be an ambiguous situation in the event that someone actually wanted to compare by reference.

### DIFF
--- a/docs/rules/no-incorrect-deep-equal.md
+++ b/docs/rules/no-incorrect-deep-equal.md
@@ -1,6 +1,6 @@
-# Avoid using `deepEqual` with literals
+# Avoid using `deepEqual` with primitives
 
-`deepEqual` and `notDeepEqual` are unnecessary when comparing literals, template strings or `undefined`. Use `is` or `not` instead.
+The `deepEqual` and `notDeepEqual` assertions are unnecessary when comparing primitives. Use `is` or `not` instead.
 
 This rule is fixable.
 
--- a/docs/rules/no-incorrect-deep-equal.md
+++ b/docs/rules/no-incorrect-deep-equal.md
@@ -1,0 +1,28 @@
+# Avoid using `deepEqual` with literals
+
+`deepEqual` and `notDeepEqual` are unnecessary when comparing literals, template strings or `undefined`. Use `is` or `not` instead.
+
+This rule is fixable.
+
+## Fail
+
+```js
+t.deepEqual(expression, 'foo');
+t.deepEqual(expression, 1);
+t.deepEqual(expression, `foo${bar}`);
+t.deepEqual(expression, null);
+t.deepEqual(expression, undefined);
+t.notDeepEqual(expression, undefined);
+```
+
+
+## Pass
+
+```js
+t.is(expression, 'foo');
+
+t.deepEqual(expression, otherExpression);
+t.deepEqual(expression, {});
+t.deepEqual(expression, []);
+t.notDeepEqual(expression, []);
+```
--- a/docs/rules/no-incorrect-deep-equal.md
+++ b/docs/rules/no-incorrect-deep-equal.md
@@ -4,6 +4,7 @@ The `deepEqual` and `notDeepEqual` assertions are unnecessary when comparing pri
 
 This rule is fixable.
 
+
 ## Fail
 
 ```js
--- a/readme.md
+++ b/readme.md
@@ -46,6 +46,7 @@ Configure it in `package.json`.
 			"ava/no-identical-title": "error",
 			"ava/no-ignored-test-files": "error",
 			"ava/no-import-test-files": "error",
+			"ava/no-incorrect-deep-equal": "error",
 			"ava/no-invalid-end": "error",
 			"ava/no-nested-tests": "error",
 			"ava/no-only-test": "error",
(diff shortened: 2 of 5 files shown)


docs/rules/no-incorrect-deep-equal.md
```diff
@@ -0,0 +1,28 @@
+# Avoid using `deepEqual` with literals
+
+`deepEqual` and `notDeepEqual` are unnecessary when comparing literals, template strings or `undefined`. Use `is` or `not` instead.
```
sindresorhus: Primitives is a more correct word than literals (there could be object literal).

```suggestion
The `deepEqual` and `notDeepEqual` assertions are unnecessary when comparing primitives. Use `is` or `not` instead.
```

docs/rules/no-incorrect-deep-equal.md
```diff
@@ -0,0 +1,28 @@
+# Avoid using `deepEqual` with literals
```
sindresorhus:
```suggestion
# Avoid using `deepEqual` with primitives
```

sindresorhus:
> Does not cover t.deepEqual(1, expression); // ? situations. 

I think it should handle that, as long as the first or second argument is a literal primitive.

> Does not fix back to deepEqual if it detects a non-literal in is -- I felt that would be an ambiguous situation in the event that someone actually wanted to compare by reference.

👍

test/no-incorrect-deep-equal.js
```diff
@@ -0,0 +1,192 @@
...
+			test(t => {
+				t.notDeepEqual(expression, []);
+			});
+		`
+	],
+	invalid: [
+		{
+			code: `
+				${header}
+				test(t => {
```
sindresorhus: I know this doesn't matter here, but AVA requires tests to have a title, so would be good to do that. Can just be `test('x', ..`.

readme.md
```diff
@@ -83,6 +84,7 @@ The rules will only activate in test files.
 - [no-identical-title](docs/rules/no-identical-title.md) - Ensure no tests have the same title.
 - [no-ignored-test-files](docs/rules/no-ignored-test-files.md) - Ensure no tests are written in ignored files.
 - [no-import-test-files](docs/rules/no-import-test-files.md) - Ensure no test files are imported anywhere.
+- [no-incorrect-deep-equal](docs/rules/no-incorrect-deep-equal.md) - Avoid using `deepEqual` with literals. *(fixable)*
```
sindresorhus: The description should be in sync with the rule docs heading.

rules/no-incorrect-deep-equal.js
```diff
@@ -0,0 +1,125 @@
...
+
+module.exports = {
+	create,
+	meta: {
+		docs: {
+			url: util.getDocsUrl(__filename)
+		},
+		fixable: true,
+		messages: {
+			[MESSAGE_ID_LITERAL]: 'Avoid using `{{callee}}` with literals',
```
sindresorhus:
```suggestion
			[MESSAGE_ID_LITERAL]: 'Avoid using `{{callee}}` with literal primitives',
```

And I don't think we need the following two messages.

test/no-incorrect-deep-equal.js
```diff
@@ -0,0 +1,312 @@
...
+	messageId: 'no-deep-equal-with-undefined'
+};
+
+const header = 'const test = require(\'ava\');\n';
+
+ruleTester.run('no-incorrect-deep-equal', rule, {
+	valid: [
+		`
+			${header}
+			test(t => {
```
sindresorhus: Add `'x'` titles for all the tests here.

sindresorhus:
@MrHen Nice work as usual 👌

PR #265, opened 2019-06-15
Add `hooks-order` rule

### LINKED ISSUE #196: Rule proposal: Enforce hooks to be at the top of the test file
[Issuehunt badges](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/196)

It's useful to be able to see the hooks first for readability.

Proposed order: `before`, `after`, `beforeEach`, `afterEach`, `test`

Thoughts?

> There is a $60.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/196).

MrHen:
Closes #196.

### DIFF
--- a/rules/hooks-order.js
+++ b/rules/hooks-order.js
@@ -1,0 +1,114 @@
+'use strict';
+const {visitIf} = require('enhance-visitors');
+const createAvaRule = require('../create-ava-rule');
+const util = require('../util');
+
+const MESSAGE_ID = 'hooks-order';
+
+const create = context => {
+	const ava = createAvaRule();
+
+	let before = null;
+	let after = null;
+	let beforeEach = null;
+	let afterEach = null;
+	let test = null;
+
+	return ava.merge({
+		'CallExpression[callee.object.name="test"][callee.property.name="before"]': visitIf([
+			ava.isInTestFile,
+			ava.isTestNode
+		])(node => {
+			before = node.callee.property.name;
+
+			const invalidNode = after || beforeEach || afterEach || test;
+
+			if (invalidNode) {
+				context.report({
+					node,
+					messageId: MESSAGE_ID,
+					data: {
+						current: before,
+						invalid: invalidNode
+					}
+				});
+			}
+		}),
+		'CallExpression[callee.object.name="test"][callee.property.name="after"]': visitIf([
+			ava.isInTestFile,
+			ava.isTestNode
+		])(node => {
+			after = node.callee.property.name;
+
+			const invalidNode = beforeEach || afterEach || test;
+
+			if (invalidNode) {
+				context.report({
+					node,
+					messageId: MESSAGE_ID,
+					data: {
+						current: after,
+						invalid: invalidNode
+					}
+				});
+			}
+		}),
+		'CallExpression[callee.object.name="test"][callee.property.name="beforeEach"]': visitIf([
+			ava.isInTestFile,
+			ava.isTestNode
+		])(node => {
+			beforeEach = node.callee.property.name;
+
+			const invalidNode = afterEach || test;
+
+			if (invalidNode) {
+				context.report({
+					node,
+					messageId: MESSAGE_ID,
+					data: {
+						current: beforeEach,
+						invalid: invalidNode
+					}
+				});
+			}
+		}),
+		'CallExpression[callee.object.name="test"][callee.property.name="afterEach"]': visitIf([
+			ava.isInTestFile,
+			ava.isTestNode
+		])(node => {
+			afterEach = node.callee.property.name;
+
+			const invalidName = test;
+
+			if (invalidName) {
+				context.report({
+					node,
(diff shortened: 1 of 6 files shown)


test/hooks-order.js
```diff
@@ -0,0 +1,277 @@
...
+	env: {
+		es6: true
+	}
+});
+
+const errors = [{ruleId: 'no-todo-test'}];
+const header = 'const test = require(\'ava\');\n';
+
+ruleTester.run('no-todo-test', rule, {
+	valid: [
```
GMartigny: You could add more valid scenarii (no header, just one test block ...) to protect from future regressions.

test/hooks-order.js
```diff
@@ -0,0 +1,277 @@
+
+import test from 'ava';
+import avaRuleTester from 'eslint-ava-rule-tester';
+import {outdent} from 'outdent';
```
GMartigny: Why bother with a new dependency ? Additional indent is not an issue here.

MrHen: It's not necessary. I was copying a pattern from a different repo.

test/hooks-order.js
```diff
@@ -0,0 +1,390 @@
...
+			test.afterEach.always((t) => {
+				doFoo();
+			});
+
+			test('foo', (t) => {
+				t.true(true);
+			});
+		`,
+		`
+			test.before((t) => {
```
GMartigny: The point of "headless" tests is to validate that the rule don't report an error on non-test files. It should be an invalid case, not reported because out of a test file.

MrHen: I don't understand what you mean. Do you want me to move it to the `invalid` section but with no `errors`?

I assumed the `valid` / `invalid` areas of the test file were named that because of how the test worked.

GMartigny: I'm sorry if it wasn't clear, I mean that rules should not report errors outside of test file. Those test files are detected by the presence of `import test from 'ava';`.

A seemingly invalid test-case without this header should not fail and therefore be valid. (e.g [`max-asserts`](https://github.com/avajs/eslint-plugin-ava/blob/master/test/max-asserts.js#L40))

MrHen: Isn't that what this is doing? The case you are commenting on will not report a lint error for the exact reason you state.

GMartigny: Yep, however the point is to write an invalid test-case to check that it didn't report because it miss a "header" and not because it's a valid test-case.

This is valid:
```js
// No header, wrong order
test.after(t => t.true(1 === 1));

test.before(t => t.true(1 === 1));
```

MrHen: Ah, gotcha. That's exact case is on [line 88](https://github.com/avajs/eslint-plugin-ava/pull/265/files/510784dc44c71189691d05f8d34dd8eb0958d322#diff-24bb544a83a06a413ab2ad3d665bfdcaR88):

```
		`
 			test.after((t) => {
 				doFoo();
 			});

 			test.before((t) => {
 				doFoo();
 			});
 		`,
```

sindresorhus:
Can you make it auto-fixable? I think we can only make it auto-fixable when there are no other variables or function between the test calls, otherwise it's not guaranteed to be safe.

test/hooks-order.js
```diff
@@ -0,0 +1,390 @@
...
+			});
+
+			test.after.always((t) => {
+				doFoo();
+			});
+
+			test.beforeEach((t) => {
+				doFoo();
+			});
+
```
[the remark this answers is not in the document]
MrHen: What behavior are you expecting? I'll add some tests covering the scenario but I would still expect the same things to get flagged.

rules/hooks-order.js
```diff
@@ -0,0 +1,154 @@
...
+
+	let before = null;
+	let after = null;
+	let afterAlways = null;
+	let beforeEach = null;
+	let afterEach = null;
+	let afterEachAlways = null;
+	let test = null;
+
+	return ava.merge({
```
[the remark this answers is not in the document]
MrHen: I cleaned it up a bit. There are a few other ways this could get DRY'd up more but I'm not sure how far down that path I really want to go.

MrHen:
@sindresorhus Added a fixer. I made a few calls around whitespace and comments during fixing:

```js
				test.after((t) => {
					doFoo();
				});

				// comments
				test.before((t) => {
					doFoo();
				});
```

Gets fixed to:

```js
				// comments
				test.before((t) => {
					doFoo();
				});

				test.after((t) => {
					doFoo();
				});
```

Anything other than whitespace and comments between hooks will prevent the fixer from activating.

test/hooks-order.js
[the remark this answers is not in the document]
sindresorhus: Yup, just wanted to make sure it's tested.

sindresorhus:
Nice work, @MrHen ✨

PR #266, opened 2019-06-17
Add missing period in some rule messages

JLHwung:
I've done a grep and it turns out these two rules have missing period in the report message. Now I add them for consistency.


sindresorhus:
You need to update their tests too, and please don't use `[ci skip]` for actual code changes.

JLHwung:
> You need to update their tests too, and please don't use `[ci skip]` for actual code changes.

Ahh, my bad. I have updated tests and pushed.

PR #283, opened 2019-12-24
docs: fix examples for `test-title-format`

gurpreetatwal:
The examples incorrect enabled the `test-title` rule instead of the `test-title-format` rule


novemberborn:
Thanks @gurpreetatwal!

PR #286, opened 2020-01-02
Fix integration test

fisker:

### DIFF
--- a/test/integration/eslint-config-ava-tester/package-lock.json
+++ b/test/integration/eslint-config-ava-tester/package-lock.json
@@ -4,49 +4,48 @@
     "lockfileVersion": 1,
     "dependencies": {
         "@babel/code-frame": {
-            "version": "7.0.0",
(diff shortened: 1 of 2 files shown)


sindresorhus:
Thanks for fixing 👌

fisker:
Master not running, says `Build type disabled via repository settings (push).` https://travis-ci.org/avajs/eslint-plugin-ava/requests

Is it because of this line? https://github.com/avajs/eslint-plugin-ava/blob/master/.travis.yml#L12

maybe 

```diff
-  - if [ $INTEGRATION == true ]; then npm run integration; fi
+  - if [[ $INTEGRATION == true ]]; then npm run integration; fi
```

sindresorhus:
Fixed: https://travis-ci.org/avajs/eslint-plugin-ava/jobs/631839007 Somehow the Travis setting to build on branches was disabled.

> -  - if [ $INTEGRATION == true ]; then npm run integration; fi
+  - if [[ $INTEGRATION == true ]]; then npm run integration; fi

We should still do this though. As `[[` is better and to be as close as possible to the Unicorn setup.

PR #287, opened 2020-01-03
Update `.travis.yml`; don't test Node.js 6 & 8

fisker:
Drop node.js 6?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
 language: node_js
 node_js:
+  - '13'
   - '12'
   - '10'
-  - '8'
-  - '6'
 matrix:
   include:
     - node_js: '12'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,15 @@
 language: node_js
 node_js:
-  - 12
-  - 10
-  - 8
-  - 6
+  - '12'
+  - '10'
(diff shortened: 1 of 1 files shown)


novemberborn:
We can drop Node.js 6 and 8, actually. And add 13.

If you're keen perhaps you could change this to use GitHub Actions? Have a look at https://github.com/avajs/ava/blob/master/.github/workflows/ci.yml for inspiration.

fisker:
@novemberborn updated.

> Not really, unless it solves some huge problems. GitHub Actions is still very early and I don't want to deal with issues with it. The config is also very verbose compared to Travis.

https://github.com/sindresorhus/eslint-plugin-unicorn/issues/361#issuecomment-528988930

novemberborn:
> > Not really, unless it solves some huge problems. GitHub Actions is still very early and I don't want to deal with issues with it. The config is also very verbose compared to Travis.
> 
> [sindresorhus/eslint-plugin-unicorn#361 (comment)](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/361#issuecomment-528988930)

I've merged this PR — but going forward we're using GitHub Actions for `ava` and `@ava/babel` so I'm more than happy to use it here as well.

PR #294, opened 2020-02-15
Allow importing AVA as `anyTest` in TypeScript files

### LINKED ISSUE #293: Rule `use-test` conflicting with typing of `t.context`
The [AVA documentation](https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md#typing-tcontext) recommends to to type `t.context` as follow:
```js
import anyTest, {TestInterface} from 'ava';

const test = anyTest as TestInterface<{foo: string}>;
```

However importing AVA as `anyTest` instead of `test` triggers the `use-test` rule.

I can make a PR but I don't know what the best solution is. Any suggestion?

pvdlg:
Fix #293 

Change `use-test` to allow importing AVA as `anyTest` in order to account for the example in [Typing t.context](https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md#typing-tcontext).

This change could be further improved by making sure that when AVA is imported as `any` it is assigned to a variable name `test`, e.g:
```
const test = anyTest as TestInterface<{foo: string}>;
```

In other words the following should be invalid:
```
import anyTest, {TestInterface} from 'ava';

const myTest = anyTest as TestInterface<{foo: string}>;
```

That's important because some other rules depends on the AVA object to be named `test`. However, the change in that PR is still better than forcing users to disable the rules.

Implementing such validation is beyond my knowledge of ESLint aPI though...

### DIFF
--- a/docs/rules/use-test.md
+++ b/docs/rules/use-test.md
@@ -3,7 +3,7 @@
 Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/related/eslint-plugin-ava/docs/rules/use-test.md)
 
 The convention is to import AVA and assign it to a variable named `test`. Most rules in `eslint-plugin-ava` are based on that assumption.
-In a TypeScript file (`.ts` or `.tsx`) AVA can be assign to to a variable named `anyTest` in order to define the types of `t.context` (see [Typing t.context](https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md#typing-tcontext)).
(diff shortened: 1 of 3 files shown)


novemberborn:
> Implementing such validation is beyond my knowledge of ESLint aPI though...

I've opened https://github.com/avajs/eslint-plugin-ava/issues/295 to track.

turadg:
JSDoc in `.js` files is an increasingly popular way to use TypeScript. e.g.,

```js
import anyTest from 'ava';
/** @type {import('ava').TestInterface<{ data: { kernelBundles: any, config: any } }>} */
const test = anyTest;
```

Any concerns with allowing `anyTest` in `.js` too? (i.e. removing the filename constraint)

PR #301, opened 2020-05-01
Allow three snapshot arguments

BarryThePenguin:
According to the AVA [documentation for `.snapshot()`](https://github.com/avajs/ava/blob/master/docs/03-assertions.md#snapshotexpected-options-message) it can accept three argument

`.snapshot(expected, options?, message?)`

The current lint rules don't allow using `.snapshot()` with the third argument. This PR updates the rule so the third argument is allowed

One thing I noticed putting this together, I had to change the number of arguments in some of the test assertions as they were failing. I'm not sure if this is correct, but it appears the same is done for `.throws()`? Happy to make any changes

### DIFF
--- a/rules/assertion-arguments.js
+++ b/rules/assertion-arguments.js
@@ -71,7 +71,7 @@ const expectedNbArguments = {
 	},
 	snapshot: {
 		min: 1,
-		max: 2
+		max: 3
 	},
 	teardown: {
 		min: 1,
--- a/test/assertion-arguments.js
+++ b/test/assertion-arguments.js
@@ -178,6 +178,7 @@ ruleTester.run('assertion-arguments', rule, {
 		testCase(false, 't.true(true, \'message\');'),
 		testCase(false, 't.truthy(\'unicorn\', \'message\');'),
 		testCase(false, 't.snapshot(value, \'message\');'),
+		testCase(false, 't.snapshot(value, options, \'message\');'),
 		testCase(false, 't.context.plan();'),
 		testCase(false, 't.teardown(() => {});'),
 		testCase(false, 't.timeout(100, \'message\');'),
@@ -209,6 +210,7 @@ ruleTester.run('assertion-arguments', rule, {
 		testCase(false, 't.true(true);'),
 		testCase(false, 't.truthy(\'unicorn\');'),
 		testCase(false, 't.snapshot(value);'),
+		testCase(false, 't.snapshot(value, options);'),
 		// Shouldn't be triggered since it's not a test file
 		testCase(false, 't.true(true, \'message\');', [], {useHeader: false}),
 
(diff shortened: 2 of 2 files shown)


novemberborn:
> I had to change the number of arguments in some of the test assertions as they were failing. I'm not sure if this is correct, but it appears the same is done for `.throws()`? Happy to make any changes

I think this is because the rule has an option to require an assertion message. Which is assumed to be the last argument. But this PR changes that to be the third argument, instead of the second. Which technically is a breaking change. It's also really awkward, you'd now have to write `t.snapshot(expected, undefined, 'message')` rather than `t.snapshot(expected, 'message')`.

@sindresorhus can we change the heuristic for whether a message is passed to look for the last string argument? And if it's a variable, just assume it's a message?

https://github.com/avajs/eslint-plugin-ava/blob/404e742c755589e2c383920e7a6c53eccd7a0b30/rules/assertion-arguments.js#L141

sindresorhus:
> @sindresorhus can we change the heuristic for whether a message is passed to look for the last string argument? And if it's a variable, just assume it's a message?

Wouldn't this mean that `t.is(foo, 'x');` would be assumed to be the message? Which would not be great. Or do you mean that it would look at the last argument as long as it's after the minimum number of arguments?

novemberborn:
I think this would only apply to `t.snapshot()`, because it is either `t.snapshot(expected, options, message)` or `t.snapshot(expected, message)`.

sindresorhus:
Then I would just special-case `t.snapshot`.

sindresorhus:
Bump

BarryThePenguin:
Just checking that there's nothing waiting on me? I may need some guidance if there's any other changes

novemberborn:
@BarryThePenguin I believe with these changes, this would be invalid:

```js
t.snapshot(expected, 'should match expectations')
```

All of these should be valid:

```js
t.snapshot(expected, 'should match expectations')
t.snapshot(expected, { id: 'expected' }, 'should match expectations')
t.snapshot(expected, undefined, 'should match expectations')
```

So while the maximum number of arguments should indeed be `3`, the `assertion-arguments` rule needs a special case to recognize that first scenario.

sindresorhus:
@BarryThePenguin Don't force push. That makes it impossible to see what changed since the last comment.

sindresorhus:
> So while the maximum number of arguments should indeed be 3, the assertion-arguments rule needs a special case to recognize that first scenario.

@BarryThePenguin ^

novemberborn:
Sorry this dragged out so much @BarryThePenguin. As you might have realized AVA 4 will no longer take an options argument for its snapshot assertion, so these changes are no longer needed. Thank you for your work on this.

BarryThePenguin:
No worries! I was a bit overwhelmed by all the pull requests I had open, so did a bit of cleaning..

I did give this one another attempt, but wasn't able to quite get my head around the tests for assertions. At least to ensure I was covering all the cases..

I'm still using, and loving AVA 😊 Tanks! ❤️

PR #310, opened 2020-08-15
Upgrade to ESLint 7

novemberborn:
Drop ESLint 6 support.

I, uhm, kinda brute forced things so that the tests would pass. I don't actually understand what I've changed.


sindresorhus:
We did a similar migration here too: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/577

novemberborn:
Thanks @sindresorhus

PR #314, opened 2020-12-04
Ensure the assertion message is a string

### LINKED ISSUE #167: The `assertion-arguments` rule should validate message argument type
[Issuehunt badges](https://issuehunt.io/r/avajs/eslint-plugin-ava/issues/167)

This currently passes `t.deepEqual({}, {}, {});`, which can make for some hard to debug issues. @gajus [reported](https://gitter.im/avajs/ava?at=5880e744e836bf70106558a4) such issue on Gitter.

We should ensure that if `message` is specified, it's a string. We could implement inline literal validation first, as that's pretty easy.

We should also detect the type if it's a variable defined in the same scope:

```js
const message = 'foo';

...

t.deepEqual({}, {}, message);
```

@jfmengels Many rules could use some kind of naive type inference. Maybe [`eslint-ast-utils`](https://github.com/jfmengels/eslint-ast-utils) could add something like that?

---

> *Note: This issue has a bounty, so it's expected that you are an experienced programmer and that you give it your best effort if you intend to tackle this. Don't forget, if applicable, to add tests, docs (double-check for typos). And don't be sloppy. Review your own diff multiple times and try to find ways to improve and simplify your code. Instead of asking too many questions, present solutions. The point of an issue bounty is to reduce my workload, not give me more. Include a 🦄 in your PR description to indicate that you've read this. Thanks for helping out 🙌* - @sindresorhus

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $60.00)

### Submitted pull Requests
- [#314 Add check for the assertion message to be a string](https://issuehunt.io/r/avajs/eslint-plugin-ava/pull/314)
---

### Tips

</details>

GMartigny:
This will check for the presence of the extra argument (which should be the `message`) and verify if it's a string.
It will try to lookup variable and its value in the scope, but non-linear flow could break it.

```js
let message;
function woops () {
	message = null;
}
message = 'ok';
woops();
t.is('same', 'same', message); // <= this will not raise an error
```

Should we cover this ?

Fixes #167

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#167: The `assertion-arguments` rule should validate message argument type](https://issuehunt.io/repos/49722492/issues/167)
---
</details>

### DIFF
--- a/rules/assertion-arguments.js
+++ b/rules/assertion-arguments.js
@@ -284,10 +284,11 @@ const create = context => {
 
 				if (lastArg.type === 'Identifier') {
 					const variable = findVariable(context.getScope(), lastArg);
-					let value = null;
-					variable.references.forEach(ref => {
+					let value;
+					for (const ref of variable.references) {
 						value = ref.writeExpr || value;
-					});
+					}
+
 					lastArg = value;
 				}
 
--- a/rules/assertion-arguments.js
+++ b/rules/assertion-arguments.js
@@ -1,6 +1,6 @@
 'use strict';
 const {visitIf} = require('enhance-visitors');
-const {getStaticValue, isOpeningParenToken, isCommaToken} = require('eslint-utils');
+const {getStaticValue, isOpeningParenToken, isCommaToken, findVariable} = require('eslint-utils');
 const util = require('../util');
 const createAvaRule = require('../create-ava-rule');
 
@@ -203,6 +203,13 @@ function noComments(sourceCode, ...nodes) {
 	});
 }
 
+function isString(node) {
+	const {type} = node;
+	return type === 'TemplateLiteral' ||
+		type === 'TaggedTemplateExpression' ||
+		(type === 'Literal' && typeof node.value === 'string');
+}
+
 const create = context => {
 	const ava = createAvaRule();
 	const options = context.options[0] || {};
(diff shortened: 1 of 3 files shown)


rules/assertion-arguments.js
```diff
@@ -271,6 +278,23 @@ const create = context => {
...
 				checkArgumentOrder({node, assertion: firstNonSkipMember, context});
 			}
+
+			if (gottenArgs === nArgs.max && nArgs.min !== nArgs.max) {
+				let lastArg = node.arguments[node.arguments.length - 1];
+
+				if (lastArg.type === 'Identifier') {
+					const variable = findVariable(context.getScope(), lastArg);
+					let value = null;
+					variable.references.forEach(ref => {
```
sindresorhus, line 288: Use `Array#find` here?

rules/assertion-arguments.js
```diff
@@ -271,6 +278,23 @@ const create = context => {
 
 				checkArgumentOrder({node, assertion: firstNonSkipMember, context});
 			}
+
+			if (gottenArgs === nArgs.max && nArgs.min !== nArgs.max) {
+				let lastArg = node.arguments[node.arguments.length - 1];
+
+				if (lastArg.type === 'Identifier') {
+					const variable = findVariable(context.getScope(), lastArg);
+					let value = null;
```
sindresorhus, line 287:
```suggestion
					let value;
```

sindresorhus:
> Should we cover this ?

No. Too hard to get right and too much of an edge-case.

rules/assertion-arguments.js
GMartigny: I used `forEach` because I want to find the last writing expression. Using `find` would force me to `reverse` the array first.

```js
lastArg = variable.references.reverse().find(ref => ref.writeExpr).writeExpr;
```

Is this really better ?

sindresorhus: Shouldn't matter. The array is not large. But at least use a `for-of` loop.

PR #318, opened 2020-12-24
Don't crash on missing title

### LINKED ISSUE #317: no-identical-title crashes if title is missing
```js
/* eslint ava/no-identical-title: "error" */
const test = require('ava');
test(t => t.pass());
```
Running eslint on that results in:

```
TypeError: Cannot read property 'type' of undefined
Occurred while linting hello.ava.js:3
    at isStatic (node_modules/eslint-plugin-ava/rules/no-identical-title.js:12:31)
    at node_modules/eslint-plugin-ava/rules/no-identical-title.js:35:9
    at node_modules/enhance-visitors/index.js:25:12
    at node_modules/enhance-visitors/index.js:15:7
    at node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
```

f5lee:
If you're adding a new rule, please follow [these steps](../docs/new-rule.md).

Fixes #317

### DIFF
--- a/rules/no-identical-title.js
+++ b/rules/no-identical-title.js
@@ -31,8 +31,8 @@ const create = context => {
 			const args = node.arguments;
 			const titleNode = args.length > 1 || ava.hasTestModifier('todo') ? args[0] : undefined;
 
-			// Don't flag computed titles
-			if (!isStatic(titleNode)) {
+			// Don't flag anonymous tests or computed titles
+			if (titleNode === undefined || !isStatic(titleNode)) {
 				return;
 			}
 
--- a/test/no-identical-title.js
+++ b/test/no-identical-title.js
@@ -14,6 +14,7 @@ const header = 'const test = require(\'ava\');\n';
 
 ruleTester.run('no-identical-title', rule, {
 	valid: [
+		header + 'test(t => {});',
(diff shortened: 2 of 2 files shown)


sindresorhus:
A test without a title is not valid though.

novemberborn:
> A test without a title is not valid though.

AVA will balk at runtime, yes, but the linter rule shouldn't crash.

@f5lee thanks for the PR. Perhaps we can make the intent clearer though. First, validate that the first argument exists _and_ is a static string. Then, validate that the second argument, if any, is a function expression. If either condition is false then the rule is not applied. What do you think?

sindresorhus:
I was commenting about the test addition in the `valid` array.

novemberborn:
> I was commenting about the test addition in the `valid` array.

Fair enough. Do we have a "works but no action taken" array? If not… then `valid` is probably the place to put it. But could do with a comment maybe.

sindresorhus:
> then valid is probably the place to put it. But could do with a comment maybe.

A comment would do.

sindresorhus:
@f5lee Bump

sindresorhus:
Closing for lack of activity.

PR #321, opened 2021-01-13
Allow import statements without default specifier

### LINKED ISSUE #320: `use-test`: Allow import statements without default specifier
The `use-test` rule is triggered even if the import statement does not include a default specifier:

```ts
import type { TestInterface } from 'ava'; // "AVA should be imported as `test`."
```

Not a high priority issue I don't think, but it should be a quick fix, and ensure that `use-test` only concerns itself with `test`, not anything else.

Exodustx0:
Fixes #320

(This is my first (manual) PR, apologies in advance if I've missed anything.)


Exodustx0:
Not quite sure what's up with the integration test; it worked no errors before. Then again, it's wholly unfamiliar territory to me.

sindresorhus:
> Not quite sure what's up with the integration test

I think the error is pretty clear if you look at it.

Exodustx0:
> I think the error is pretty clear if you look at it.

Right, my bad. Wasn't paying proper attention; the issue has now been resolved however.

sindresorhus:
It would be good to add a test for that case that crashed too.

Exodustx0:
> It would be good to add a test for that case that crashed too.

Done.

sindresorhus:
Can you also update the pass/fail exampels in the docs with more things? For example, `import type ...`.

sindresorhus:
@exodustx0 Bump

sindresorhus:
@exodustx0 Bump

PR #324, opened 2021-03-18
Add an option to avoid fixing no-skip-test

edbrannin:
My team's projects run `eslint --fix` by deafult, and some of our editors apply lint-fixes on save by default.

This has led to some issues with this rule (with I 100% wish to keep using):
* People adding `// eslint-disable-next-line ava/no-skip-test` to get around their editors' auto-fix, then accidentally committing the whole file without fixing it
* Adding this rule to an existing project can strip `test.skip()` in a bunch of places that deserve manual inspection

This option also lets projects opt-in to working like the [eslint docs recommend](https://eslint.org/docs/developer-guide/working-with-rules#applying-fixes):

> Best practices for fixes:
>
> 1. Avoid any fixes that could change the runtime behavior of code and cause it to stop working.

If this is accepted, I'd like to do the same thing for `no-only-test` for the same reasons.

Issue #281, opened 2019-11-23
Support the ESLint Suggestions API


sindresorhus:
There have been many cases where we could not provide auto-fixers because it was either not safe or there were multiple possible fixes. The [Suggestions API](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions) enables us to provide multiple fixes that can be manually chosen by a user in an editor.

We should go through existing rules with no auto-fixers or with incomplete auto-fixers and make them support the Suggestions API.

sindresorhus:
I suggest (pun intended) that we drop auto-fixer for the `no-only-test` and `no-skip-test` and instead implement suggestions for them. Auto-fixers for those can be annoying (https://github.com/SamVerschueren/vscode-linter-xo/issues/78) when you have "fix on save" in an editor.

@novemberborn Thoughts?

novemberborn:
Agreed.

PR #325, opened 2021-03-19
Use suggestions API for `no-only-test` and `no-skip-test`

edbrannin:
Closes https://github.com/avajs/eslint-plugin-ava/pull/324
https://github.com/avajs/eslint-plugin-ava/issues/281#issuecomment-587137623

### DIFF
--- a/docs/rules/no-only-test.md
+++ b/docs/rules/no-only-test.md
@@ -4,9 +4,6 @@ Translations: [Français](https://github.com/avajs/ava-docs/blob/main/fr_FR/rela
 
 It's easy to run only one test with `test.only()` and then forget about it. It's visible in the results, but still easily missed. Forgetting to remove `.only`, means only this one test in the whole file will run, and if not caught, can let serious bugs slip into your codebase.
 
-This rule is fixable. It will remove the `.only` test modifier.
-
-
 ## Fail
 
 ```js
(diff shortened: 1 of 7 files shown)


edbrannin:
@sindresorhus Integration tests are failing in the same way as they did in #324.  Was the build already broken?

sindresorhus:
You need to update the rule docs and the readme.

edbrannin:
Whoops, thanks!  I'll fix.

...actually, I checked the eslint code rules, the [only]() [two]() that support the `suggest:` API don't say anything about that or autofix in the docs.

Should I just drop references to Fixable for these rules, or would you prefer some other language?

[only]: https://eslint.org/docs/rules/no-useless-escape
[two]: https://eslint.org/docs/rules/no-unsafe-negation

sindresorhus:
> Should I just drop references to Fixable for these rules, or would you prefer some other language?

Drop

edbrannin:
Done

(oh yay, the integration test is working now too!)

edbrannin:
Thanks!

If you don't mind my asking, any idea when this might be released?  I'm very much looking forward to updating this plugin on my services.

sindresorhus:
It was released half an hour ago.

edbrannin:
Oh, double-thanks then!  Keep up the good work!

PR #341, opened 2021-11-05
Enforce strict mode

fisker:


sindresorhus:
CI is failing. Looks unrelated but weird that it fails only here and not the main branch.

fisker:
Disabled that rule and opened an issue to track #342

fisker:
Maybe because it's not strict, `eslint-plugin-eslint-plugin` can't locate something?

sindresorhus:
Seems so

PR #343, opened 2022-01-07
`use-test`: Ignore type import

### LINKED ISSUE #320: `use-test`: Allow import statements without default specifier

futpib:
This fixes #320 by ignoring type-only imports (like `import type { Macro } from 'ava';`).


sindresorhus (approved):

novemberborn:
Shipped in 13.2.0.

Issue #156, opened 2016-11-13
Rule proposal: `prefer-t-throws`


gajus:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2482%20Rewarded-%237E24E3.svg" />](https://oss.issuehunt.io/r/avajs/eslint-plugin-ava/issues/156)

I was reviewing some codes using ava and found this pattern repeating often:

```js
try {
  await request(requestOptions);
} catch (error) {
  t.true(error.statusCode === 500);
}
```

This can be written instead as:

```js
const error = await t.throws(request(requestOptions));

t.true(error.statusCode === 500);
```

I think the latter should be preferred.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $82.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://oss.issuehunt.io/u/issuehunt) ($80.00)
- [<img src='https://avatars.githubusercontent.com/u/170270?v=4' alt='sindresorhus' width=24 height=24> sindresorhus](https://oss.issuehunt.io/u/sindresorhus) ($2.00)

### Submitted pull Requests
- [#360 fix(prefer-t-throws): detect try/catch asserting on caught error](https://oss.issuehunt.io/r/avajs/eslint-plugin-ava/pull/360)
- [#835 Add `request` and `options` to `beforeError` hook state](https://oss.issuehunt.io/r/sindresorhus/ky/pull/835)
---

### Tips

- Checkout the [Issuehunt explorer](https://oss.issuehunt.io/r/avajs/eslint-plugin-ava/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://oss.issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

jfmengels:
Sounds like a good idea to me :+1:. Not sure when/why try/catch makes sense in tests when t.throws is available.

The second example is better, but FYI, using `t.fail()` would improve your first example.
```js
try {
  await request(requestOptions);
  t.fail();
} catch (error) {
  t.true(error.statusCode === 500);
}
```

gajus:
> The second example is better, but FYI, using t.fail() would improve your first example.

Another idea for a lint rule. :-)

sindresorhus:
👍 

[...]

PR #345, opened 2022-02-06
Add `prefer-t-throws` rule

Mesteery:
Fixes: https://github.com/avajs/eslint-plugin-ava/issues/156


docs/rules/prefer-t-throws.md
```diff
@@ -0,0 +1,60 @@
...
+This rule will enforce the use of `t.throws()` when possible.
+
+## Fail
+
+```js
+const test = require('ava');
+
+test('some test', async t => {
+	try {
+		await Promise.reject('error');
```
sindresorhus, line 12:
```suggestion
		await Promise.reject('error');
```

This will confuse users. I know it's just an example, but it makes no sense.

sindresorhus, line 12:
```suggestion
		await potentiallyThrowingFunction();
```

docs/rules/prefer-t-throws.md
```diff
@@ -0,0 +1,60 @@
...
+		t.fail();
+	} catch (error) {
+		t.is(error, 'error');
```
sindresorhus, line 15:
```suggestion
		t.is(error.message, 'Unicorn overload');
```

docs/rules/prefer-t-throws.md
```diff
@@ -0,0 +1,60 @@
...
+		await promiseThatMaybeFails();
+		await anotherPromise;
+		await timeout(100, 'good error message');
+		t.fail();
+	} catch (error) {
+		t.is(error, 'good error message');
+	}
+});
+```
+
```
sindresorhus, line 60:
```suggestion
```

sindresorhus:
AVA now has `t.throws` and `t.throwsAsync`. See the AVA docs.

test/prefer-t-throws.js
```diff
@@ -0,0 +1,27 @@
...
+		`${header}test(async t => { try { await promise; } catch (error) { t.is(error, 'error'); } });`,
+		`${header}test(async t => { try { await promise; await anotherPromise; t.fail(); } catch (error) { t.is(error, 'error'); } });`,
+	],
+	invalid: [
+		{
+			code: `${header}test(async t => { try { await Promise.reject('error'); t.fail(); } catch (error) { t.is(error, 'error'); } });`,
+			errors: [{message: 'Prefer using the `t.throws()` assertion.'}],
+		},
+	],
+});
```
sindresorhus, line 27: I'd like to see more tests. Try to think of edge-cases.

Mesteery:
> AVA now has t.throws and t.throwsAsync. See the AVA docs.

So should the rule be `t-throws-async` instead or should I add another rule?

sindresorhus:
The name should be the what it is, but you need to also support the async method.

rules/prefer-t-throws.js
```diff
@@ -0,0 +1,146 @@
+'use strict';
+
+const {visitIf} = require('enhance-visitors');
+const createAvaRule = require('../create-ava-rule');
+const util = require('../util');
+
+function hasAwaitExpression(nodes) {
```
sindresorhus, line 7: There must be a simpler way to achieve this.

Mesteery: I don't know, but the function does a bit more than its name indicates: it looks for an AwaitExpression, which is not in a (arrow) function, and throws if it meets a return.

sindresorhus: At least add a code comments that succinctly explains what the function does.

docs/rules/prefer-t-throws.md
```diff
@@ -0,0 +1,96 @@
...
+## Pass
+
+```js
+const test = require('ava');
+
+test('some test', async t => {
+	const error = await t.throwsAsync(asyncThrowingFunction());
+	t.is(error.message, 'Unicorn overload');
```
sindresorhus, line 55: This is not the recommended way to test the error message.

rules/prefer-t-throws.js
```diff
@@ -0,0 +1,149 @@
+'use strict';
+
+const {visitIf} = require('enhance-visitors');
+const createAvaRule = require('../create-ava-rule');
+const util = require('../util');
+
+// This function checks if there is an AwaitExpression, which is not inside another function.
```
sindresorhus, line 7: Maybe you could find an `AwaitExpression`, then iterate upwards through its parent scopes until you encounter a function, and then check the function.

https://eslint.org/docs/developer-guide/scope-manager-interface

docs/rules/prefer-t-throws.md
```diff
@@ -0,0 +1,96 @@
...
+});
+```
+
+```js
+const test = require('ava');
+
+test('some test', async t => {
+	try {
+		const result = await potentiallyThrowingFunction();
+		// This passes because using a try-catch can be intentional, for example, to test both the result and the error:
```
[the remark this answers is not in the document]
Mesteery: Indeed, I can't remember what was in my head.

rules/prefer-t-throws.js
Mesteery: What about the `ReturnStatement`? I doubt there's anything simpler than that in the end.

docs/rules/prefer-t-throws.md
sindresorhus: That is also not the recommended way. Please read the docs for `t.throwsAsync`. It accepts a `message` option.

rules/prefer-t-throws.js
sindresorhus: Do you have a realistic example of when someone would actually use `return` in a try/catch when testing an error?

Mesteery: No, I don't have one. It's pretty hard to find a realistic example but I have a feeling that there might be some code written that way.

sindresorhus: > I have a feeling that there might be some code written that way.

Rule logic should be based on test fixtures, not feelings. My recommendation to simplify the logic here stands.

Issue #347, opened 2022-08-15
`prefer-t-regex` fails with `t.is()` with a single argument


ehmicky:
When `t.is()` is passed 0 or 1 argument instead of 2, `prefer-t-regex` crashes. 

This is related to https://github.com/avajs/eslint-plugin-ava/issues/346, but that other issue seems to be more focused on graceful failures in general.

## Steps to reproduce

`test.js`:

```js
import test from 'ava'

test('example', t => {
  t.is(true)
}) 
```

`.eslintrc.yml`:

```yml
parserOptions:
  sourceType: module
  ecmaVersion: 2022
plugins: [ava]
rules:
  ava/prefer-t-regex: 2
```

```
$ eslint test.js 

Oops! Something went wrong! :(

ESLint: 8.22.0

TypeError: Cannot read properties of undefined (reading 'regex')
Occurred while linting /home/ehmicky/Desktop/test.js:4
Rule: "ava/prefer-t-regex"
    at isRegExp (/home/ehmicky/Desktop/node_modules/eslint-plugin-ava/rules/prefer-t-regex.js:83:14)
    at equalityHandler (/home/ehmicky/Desktop/node_modules/eslint-plugin-ava/rules/prefer-t-regex.js:138:32)
    at /home/ehmicky/Desktop/node_modules/eslint-plugin-ava/rules/prefer-t-regex.js:195:5
    at /home/ehmicky/Desktop/node_modules/enhance-visitors/index.js:25:12
    at /home/ehmicky/Desktop/node_modules/enhance-visitors/index.js:15:7
    at ruleErrorHandler (/home/ehmicky/Desktop/node_modules/eslint/lib/linter/linter.js:1114:28)
    at /home/ehmicky/Desktop/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
    at Object.emit (/home/ehmicky/Desktop/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/home/ehmicky/Desktop/node_modules/eslint/lib/linter/node-event-generator.js:297:26)
```

ESLint: `8.22.0`
`eslint-plugin-ava`: `13.2.0`
Ava: `4.3.1`

fregante:
Duplicate of https://github.com/avajs/eslint-plugin-ava/issues/346

Issue #346, opened 2022-08-03
Some rules do not fail gracefully: ava/prefer-t-regex


spence-s:
This particular plugin causes eslint to crash with type errors for some rules if the syntax is not exactly right. Even if the document/file is parsable.

`prefer-t-regex` is the biggest complainer afaict. 

`Cannot read properties of undefined (reading 'type')`

This is super noticeable when using the vscode xo extension as we alert on an eslint crash like that, so we alert on every key stroke until the plugin can lint correctly again. 

Something like `t.anyAssertion()` causes the error.

ehmicky:
Related: https://github.com/avajs/eslint-plugin-ava/issues/347

PR #350, opened 2023-01-16
fix: guard against type errors in prefer-t-regex

spence-s:
fixes some of the prefer-t-regex issues. #346 #347

### DIFF
--- a/rules/prefer-t-regex.js
+++ b/rules/prefer-t-regex.js
@@ -51,6 +51,10 @@ const create = context => {
 	```
 	*/
 	const findRootReference = node => {
+		if (!node) {
+			return;
+		}
+
 		if (node.type === 'Identifier') {
 			const reference = findReference(node.name);
 
@@ -80,17 +84,26 @@ const create = context => {
 	2. `RegExp` class can't be looked up so the function just checks for the name `RegExp`.
 	*/
 	const isRegExp = lookup => {
-		if (lookup.regex) {
+		if (lookup?.regex) {
 			return true;
 		}
 
 		// Look up references in case it's a variable or RegExp declaration.
 		const reference = findRootReference(lookup);
+
+		if (!reference) {
+			return false;
+		}
+
 		return reference.regex ?? reference.name === 'RegExp';
 	};
 
 	const booleanHandler = node => {
 		const firstArg = node.arguments[0];
+
+		if (!firstArg) {
+			return;
+		}
 
 		const isFunctionCall = firstArg.type === 'CallExpression';
 		if (!isFunctionCall || !firstArg.callee.property) {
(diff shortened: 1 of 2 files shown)


rules/prefer-t-regex.js
```diff
@@ -51,7 +51,7 @@ const create = context => {
 	```
 	*/
 	const findRootReference = node => {
```
sindresorhus, line 53: Seems like it would be better to just guard the `node` at the top here instead.

rules/prefer-t-regex.js
```diff
@@ -156,7 +156,7 @@ const create = context => {
 		};
 
 		// Only fix a statically verifiable equality.
-		if (regex && matchee.type === 'Literal') {
+		if (regex && matchee?.type === 'Literal') {
```
sindresorhus, line 159: If `matches` is undefined, it would be clearer to guard against it right after it's defined.

Same applies to pretty much all the changes.

spence-s:
Updated for @sindresorhus preferences. 

Note the added test cases that currently cause this rule to throw 😅.

rules/prefer-t-regex.js
```diff
@@ -80,18 +84,27 @@ const create = context => {
...
-		if (lookup.regex) {
+		if (lookup?.regex) {
 			return true;
 		}
 
 		// Look up references in case it's a variable or RegExp declaration.
 		const reference = findRootReference(lookup);
+
+		if (!reference) {
+			return;
```
sindresorhus, line 95:
```suggestion
			return false;
```

rules/prefer-t-regex.js
```diff
@@ -80,18 +84,27 @@ const create = context => {
 	2. `RegExp` class can't be looked up so the function just checks for the name `RegExp`.
 	*/
 	const isRegExp = lookup => {
-		if (lookup.regex) {
```
sindresorhus, line 83: Could use a guard for `lookup` above here.

rules/prefer-t-regex.js
```diff
@@ -179,7 +197,12 @@ const create = context => {
 		CallExpression: visitIf([
 			ava.isInTestFile,
 			ava.isInTestNode,
-		])(node => {
+		],
+		)(node => {
+			if (!node || !node.callee || !node.callee.property) {
```
sindresorhus, line 202:
```suggestion
			if (!node?.callee?.property) {
```

sindresorhus (changes requested):

rules/prefer-t-regex.js
spence-s: already added 😄

sindresorhus:
I think you misunderstood me in https://github.com/avajs/eslint-plugin-ava/pull/350/files#r1072500429. I meant `if (!lookup) { return false }`. Because the current check will allow undefined to continue in the function.

spence-s:
yes - ha did misunderstand, but good call!

PR #358, opened 2024-05-04
Remove incorrect flat "plugins" export

TylerVigario:


index.js
```diff
@@ -55,9 +55,6 @@ module.exports = {
 			},
 		},
 		'flat/recommended': {
-			plugins: {
-				ava: 'ava',
-			},
```
sindresorhus, line 60: It should not be removed, just correctly defined: https://eslint.org/blog/2022/08/new-config-system-part-2/#more-powerful-and-configurable-plugins

TylerVigario: It works when removed, but it doesn't work in its current state. This is all the effort I have to commit to these changes.

PR #361, opened 2026-05-15
Update assertion-arguments rule to handle formatAsCodeBlock snapshot argument

novemberborn:
See https://github.com/avajs/ava/pull/3449.


sindresorhus:
This seeems to regress existing `t.snapshot()` message handling. The new logic treats any second argument that does not look syntactically like a string as the options object, so message variables are misread as options.

I manually checked these on the PR branch:

- `{message: 'always'}` reports `missing-message` for `const message = 'ok'; t.snapshot(value, message);`
- `{message: 'never'}` reports nothing for the same cal
- `t.snapshot(value, true)` and `t.snapshot(value, {formatAsCodeBlock: true}, true)` also report nothing

could this detect the options form more narrowly, then run the existing message validation on the real message slot? This probaly needs tests for variable messages and non-string messages.

PR #362, opened 2026-05-19
Refactor to get rid of outdated `enhance-visitors` dependency

andreww2012:
Hello!

I noticed that `enhance-visitors` package of this package:
- Last updated 10 years ago;
- Brings barely maintained & heavy `lodash` dependency, which recently received a security vulnerability;
- Quite trivial;
- Virtually [used by only 2 ESLint plugins](https://www.npmjs.com/package/enhance-visitors?activeTab=dependents).

In this PR, I decided to refactor the code to remove the dependency on it:
- `visitIf` wrapper calls was replaced with native control flow expressions;
- `mergeVisitors` implementation was inlined.

Along with that, I've slightly improve the architecture of the rules creator to get rid of the `currentTestNode` flag.

### DIFF
--- a/create-ava-rule.js
+++ b/create-ava-rule.js
@@ -1,9 +1,7 @@
-import enhance from 'enhance-visitors';
 import {getTestModifiers, unwrapTypeExpression} from './util.js';
 
-export default () => {
+export default context => {
 	let isTestFile = false;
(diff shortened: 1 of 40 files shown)


andreww2012:
Thank you :)

